Issue #6, opened 2023-01-05
Component - consulting benefits


wicksipedia:
Data sourced from consulting markdown file:
- body text
- benefits list
- rules link
- background image (defaults to the image displayed if not present)

![Image](https://user-images.githubusercontent.com/600044/211243546-0e1132b0-f11e-434d-bda2-2f51dfa7bce7.png)

PR #40, opened 2023-01-16
6 consulting benefits

johnxussw:
#6 

- [x] create component: benefit card
    - [x] scroll animation
    - [x] style
- [x] create component: benefit list
    - [x] style
    - [x] responsive
- [x] create component: rules


pages/_app.tsx
```diff
@@ -5,8 +5,9 @@ import { Analytics } from "../components/layout/analytics";
 import { NEXT_SEO_DEFAULT } from "../next-seo.config";
 import * as gtag from "../lib/gtag";
 import "aos/dist/aos.css";
-import "../styles.css";
 import AOS from "aos";
+import "aos/dist/aos.css"
```
wicksipedia, line 9: looks like a double import
```suggestion
```

plugins/normalize.js
```diff
@@ -0,0 +1,9 @@
+const plugin = require("tailwindcss/plugin");
+
+module.exports = plugin(function ({ addBase }) {
+  addBase({
```
wicksipedia, line 4: I like this but I think it should be in the tailwind.config.js file.... similar to https://play.tailwindcss.com/uF9rN9jewc?file=config

tailwind.config.js
```diff
@@ -144,6 +144,18 @@ module.exports = {
...
+              padding: "0.25rem",
+              borderRadius: "3px",
+              margin: "-0.25rem 1px",
+            },
+            "code::before": {
+              content: "",
+            },
+            "code::after": {
+              content: "",
+            },
```
wicksipedia, line 158: I don't believe this is used anywhere. It would be good to delete it then

wicksipedia (changes requested):

plugins/normalize.js
johnxussw: Done - plugin removed

Base font has been defined in style.css

tailwind.config.js
johnxussw: Done - unused code removed

pages/_app.tsx
johnxussw: Done - duplicated import removed

wicksipedia (approved):
lgtm

PR #87, opened 2023-02-01
🐛  Fix home page

bradystroud:
🫠


tkapa (approved):
Wow, seems silly that this happens with the #Megamenu

PR #125, opened 2023-02-06
💄Make upcoming events larger

### LINKED ISSUE #117: 🐛 Designer Review - Max-height the Upcoming events section
- [ ] Span the Upcoming events component to cover the full height of its section

    Image
**Figure: New site "Upcoming Events" widget**

    Image
**Figure: Current site "Upcoming Events" widget**

### LINKED ISSUE #118: 🐛 Designer Review - Reduce text line-height
- [ ]  Reduce text line-height

    Image
**Figure: New site line-height**

    Image
**Figure: Current site line-height**

tkapa:
Closes #117 
Closes #118

Note: I had issues with using `h-full` and `grow` for this. The events table would simply spill to hold all of its contents rather than just be as high as the div while allowing for scrolling.

image
_**Figure: Updated Upcoming Events section**_


pages/[filename].tsx
```diff
@@ -21,14 +21,14 @@ export default function HomePage(
       <Layout>
         <Blocks prefix="PageBeforeBody" blocks={data.page.beforeBody} />
         <Container className={"flex-1 pt-4"}>
-          <div className="gap-4 md:grid md:grid-cols-5 lg:grid-cols-3">
-            <div className="prose max-w-full md:col-span-3 lg:col-span-2">
+          <div className="gap-4 md:grid md:grid-cols-5 lg:grid-cols-5">
+            <div className="prose max-w-full leading-5 md:col-span-3 lg:col-span-3">
```
Comment (wicksipedia) [inline pages/[filename].tsx:25]: why `leading-5`? `prose` sets `line-height` by default

Comment (tkapa) [reply pages/[filename].tsx:25]: This is according to @tiagov8's feed back for #118. The default line height of `prose` is too large. I added `leading-5` to override. 

Looks like I didn't add this change in my previous PR.

I'm OK with keeping the larger line height. IMO this smaller one is ugly.

Comment (wicksipedia) [reply pages/[filename].tsx:25]: I think it should be modified in the tailwind configuration so its consistent

Comment (tkapa) [reply pages/[filename].tsx]: Strange. I removed the `leading-5` and ended up with the same line height, despite it being larger in production at the moment. Unsure why that's happened but I suppose it saves me from making more changes.

wicksipedia (approved):
lgtm - I don't think that this should be a blocker for go-live
maybe add another pbi to the backlog (down the list)

PR #158, opened 2023-02-09
59 address performance issues

johnxussw:
Performance
- [x] Create a component | an emulated Zendesk button
    - [x] Click the real Zendesk button after the script loaded
- [x] Remove duplicated analytic script and npm packages

Accessibility
- [x] Change MegaMenu main menu tag from `<a>` to `<div>`
    - [x] Change MegaMenu source code
    - [x] Change the copy of the code in this solution
- [x] Heading tags should be used in descending order without skip
- [x] Change from `<html>` to `<html lang="en">`


wicksipedia:
🐛 BUG - you have to click on the zendesk button twice to get the chat up

wicksipedia:
changes to the mega menu should be done on that repo as well - we have a copy of the code as a workaround for getting it up and running

tkapa:
Seems a shame that we aren't seeing much improvement on the Performance score :/

tkapa:
🐛 Looks like we still have to click twice on the Zendesk chat to open it up (However I think that's because of the delayed load)

components/zendeskButton/zendeskButton.tsx
```diff
@@ -0,0 +1,99 @@
...
+        clearTimeout(clickEvent.current);
+        console.log("detect >>> ", zendeskNativeButton);
+        if (zendeskNativeButton) {
+          zendeskNativeButton.click();
+        }
+      }
+
+      const repeat = () => {
+        clickButton();
+        console.log("click", maxTrials.current);
```
tkapa, line 41: This should be removed when we're happy with the behaviour.
```suggestion
```

wicksipedia:
🐛 layout issue

image

johnxussw:
> Seems a shame that we aren't seeing much improvement on the Performance score :/

True. It only improves a little bit :(
The main bundle is extremely large.

wicksipedia:
now the menu loads all the sections on page load
image

johnxussw:
We used class "unstyled" a lot in MegaMenu, but we do not define it in MegaMenu source code.
It's confusing.

johnxussw:
> now the menu loads all the sections on page load image

Done - fixed

bradystroud (approved):
lgtm ⚡️

Issue #163, opened 2023-02-09
💄 Consulting pages - Security page broken list


BrookJeynes:
- [x] The items within the "Identify Solutions" section are in a list

![image](https://user-images.githubusercontent.com/25432120/217962327-5466ef8f-d18f-4036-9869-3aed05fe5693.png)
**Figure: Items inline instead of a list**

johnxussw:
![image](https://user-images.githubusercontent.com/113034339/218912619-1ca0532e-4f42-40fc-a697-4dedb0d3b96c.png)
**Figure: https://ssw.com.au/consulting/security**

![image](https://user-images.githubusercontent.com/113034339/218912699-9aa55aa6-cb82-41ac-aef4-cb0d2cdd4f03.png)
**Figure: https://ssw.com.au/ssw/consulting/security.aspx**

PR #186, opened 2023-02-15
163 security page

johnxussw:
#163

- [x] Schema change | benefits - change description from string type to rich-text type
- [x] Replace `<span>` to  `<TinaMarkdown  />`
- [x] Change styles inside `<TinaMarkdown />`
- [x] update security.mdx file to list item

a
**Figure: list-item text edited in tina's admin site**

image
**Figure: list-item view in the description**

---

image
**Figure: emphasized text in security.mdx**

image
**Figure: emphasized text in the description**


johnxussw:
Found underscore text in the descriptions. What kind of default style we should apply to them?

- [ ] A bold
- [ ] B uses SSW red color
- [ ] C both A&B

image
**Figure: underscore text**

components/util/consulting/benefits.module.css
```diff
@@ -0,0 +1,13 @@
+.description {
```
wicksipedia, line 1: utility classes over custom css

wicksipedia (changes requested):

components/util/consulting/benefits.module.css
johnxussw: Was trying to change elements' styles inside `<TinaMarkdown />`
1. Add another prose-* in the tailwind.config.js file
2. Add custom CSS

I choose the second option since it is closer to editing in the future.

wicksipedia (approved):

PR #187, opened 2023-02-15
🐛 Seo - Fix page titles

### LINKED ISSUE #170: 🐛 Seo - Fix page titles
Page titles are inconsistent with the old site.

image
**Figure: Title inconsistency example on power bi page**
  

```[tasklist]
### Tasks
- [ ] Fix home page title
- [ ] Fix consulting page titles
- [ ] Fix 404 and other miscellaneous page titles
```

jimmidier:
Closes #170 

image
**Figure: Home page title matching the old site**

image
**Figure: Consulting page title matching the old site**


pages/consulting/[filename].tsx
```diff
@@ -50,6 +50,11 @@ export default function ConsultingPage(
     recaptchaKey: props.env["GOOGLE_RECAPTCHA_KEY"],
   };
 
+  data.consulting.seo = {
+    ...data.consulting.seo,
+    title: data.consulting.seo?.title || data.consulting.title
```
Comment (bradystroud) [inline pages/consulting/[filename].tsx:55]: Can you check where we are using `consulting.title`?

I think it could be removed now we have the title in the SEO.

If we dont need it, remove it.

bradystroud (changes requested):

pages/consulting/[filename].tsx
Comment (jimmidier) [reply pages/consulting/[filename].tsx]: Done - Removed consulting.title and used consulting.seo.title instead

bradystroud (approved):

Issue #190, opened 2023-02-16
UI - Carousel controls need to be bigger


BrookJeynes:
- [x] Enlarge carousel controls
- [x] Move carousel controls closer to each other

![image](https://user-images.githubusercontent.com/25432120/219497383-90d1c105-fb1e-4e0c-b149-bd007a003368.png)
**Figure: current website**

![image](https://user-images.githubusercontent.com/25432120/219498358-5914a648-d104-4a8a-91c8-8a83e716b887.png)
**Figure: new website**

PR #218, opened 2023-02-17
Fix - Carousel controls need to be bigger

johnxu16:
#190

- [x] Fix - enlarge carousel controls
- [x] Fix - enlarge the gap between carousel controls
- [x] Tech debt - upgrade legacy/image in the carousel


bradystroud (approved):
LGTM

PR #266, opened 2023-02-24
Megamenu hacking

wicksipedia:
To demo hacking a new megamenu UX on the website


wicksipedia:
abandoning - will restart once we are happy with a design

PR #280, opened 2023-02-27
💄Consulting Pages - Added responsive padding

### LINKED ISSUE #271: 💄Consulting pages - Inconsistent Headers Between v1 and v3
image
**FIgure: v1  /consulting/bots.aspx**

image
**Figure: v3 /consulting/bots**

Harry-Ross:
Closes #271 

On the v1 site, the line breaks were controlled by using `<br>` tags:
image
**Figure: v1  /consulting/bots.aspx**
image
**Figure: HTML controlling line breaks in above screenshot**

I added responsive padding to create greater consistency on consulting page headers, which means there does not have to be a manual entry of a break tag when implementing consistent headers across pages.

### DIFF
--- a/components/blocks/booking.tsx
+++ b/components/blocks/booking.tsx
@@ -11,7 +11,7 @@ export const Booking: FC<{
   return (
     <>
       <Container padding="px-4" className="w-full">
-        <h1 dangerouslySetInnerHTML={{ __html: title }}></h1>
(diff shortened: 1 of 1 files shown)


wicksipedia:
I like this but I think we want manual line breaks so we can put the emphasis in the right place

wicksipedia:
added line breaks to match prod with #310

PR #311, opened 2023-02-28
Update 404 page UI

### LINKED ISSUE #312: UI - Updating 404 page
```[tasklist]
### Tasks
- [ ] The 404 font is definitely not the boldest possible
See https://fonts.google.com/specimen/Open+Sans?preview.text=404&preview.text_type=custom 
I think the 800 variation of the font is not being rendered...
Can you investigate?
- [ ] Make the chain image 50% opacity (so the contrast is better when text is over it)
```

Geordie88:
updating the background so that the text will stand out when overlapping the image 
fixed loading open sans font

affected path: `/404`

closes #312 

image


tiagov8 (approved):

Geordie88 (commented):
Looks great, glad the Open sans is added

PR #313, opened 2023-02-28
Live banner - Update banner text and countdown logic

### LINKED ISSUE #289: Live banner - Update banner text and countdown logic
As per conversation with the TV team, we need to update the live banner to be:

- [ ] Add field to Events SharePoint list - LiveStreamStartDateTime
- [ ] From:
  Airing in X hours and X minutes. (6PM Sydney)
To:
  Airing in X hours and X minutes. **Live stream starts at {{LiveStartDateTime}}pm. Doors open at {{StreamStartDateTime}}pm**

jimmidier:
Closes #289 and #325

(UI checked by TV team)
image
**Figure: Updated live banner text with two timings; countdown with 30 more minutes**

Done Video at https://www.loom.com/share/1c8264b9e91a4f4fb327c4982931b810


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.22.19",
   "private": true,
   "scripts": {
-    "dev": "tinacms dev --watchFolders='components/util/' -c \"next dev\"",
+    "dev": "tinacms dev -c \"next dev\"",
```
wicksipedia, line 6: why get rid of the watch?

jimmidier: I didn't see the doc mentioning this argument anywhere, and tina dev already watches the whole project by default. Is there any particular reason that we're setting this arg and only watching the util folder?

wicksipedia: ok cool

wicksipedia:
As per conversation with @camillars and @jontygardner - hardcoding 30 minute delay is a bad idea. The event and livestream start times should be configurable

wicksipedia (changes requested):
need to reference a livestream field from sharepoint

jimmidier:
> need to reference a livestream field from sharepoint

Is there an existing field for this? The PR supports easily specifying a different delay other than the default 30 min, so I think we could merge this first if having to work on the new field.

camillars:
> > need to reference a livestream field from sharepoint
> 
> Is there an existing field for this? The PR supports easily specifying a different delay other than the default 30 min, so I think we could merge this first if having to work on the new field.

No, we need a new field in the events listing individual page https://sswcom.sharepoint.com/sites/Events/Lists/Events/AllItems.aspx (when you click on '+ new event')

image

jimmidier:
> need to reference a livestream field from sharepoint

@wicksipedia This is done. See the commit before formatting

jimmidier:
@wicksipedia As per conversation with @camillars, the new LiveStreamDelayMinutes field in SharePoint event is sufficient to support setting live stream delay minutes, as below:
image

I've also added another commit to respect StartShowBannerDateTime/EndShowBannerDateTime from SharePoint.
Please re-review

wicksipedia (approved):
lgtm

PR #326, opened 2023-03-01
Redo fix on background image

### LINKED ISSUE #302: 🐛 Consulting Template - background video is not visible
Comparing new (left) vs (old)

image

jimmidier:
Undone work
fixes #302

- [x] Used named z-index utils instead of magic numbers
- [x] Used pseudo element as video mask instead of the parent
- [x] Provided background video and siblings with proper z-indexes
- [x] Cleaned up some other existing z-indexes
- [x] A bit code formatting

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,7 +14,10 @@
   ],
   "rules": {
     "quotes": ["error", "double"],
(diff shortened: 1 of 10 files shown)


wicksipedia (approved):
lgtm - after I fixed a few linting errors

PR #405, opened 2023-03-13
⚙️Added Prettier Ignore File

Harry-Ross:
Due to the push last week where `_schema.json` was accidentally formatted with Prettier (from format on save), adding a `.prettierignore` file to prevent this from happening again.


wicksipedia (approved):
lgtm

PR #407, opened 2023-03-13
Separated schemas to separate files

### LINKED ISSUE #316: ⚙️Config - Schema File too cumbersome
Right now `config.tsx` is a very long file, would be better to move each collection to its own file and import them into `config.tsx`.

- [ ] Move each collection definition from `.tina/config.tsx` into its own file in `.tina/collections`

Harry-Ross:
Closes #316 

Added `@ts-ignore` to the config.tsx file to each schema addition - VSCode does not like the types of the schemas, however everything else runs just fine. Open to suggestions for a better way to do this.


.tina/collections/consulting.tsx
```diff
@@ -0,0 +1,218 @@
+import * as Schemas from "../../components/blocks";
+import { seoSchema } from "../../components/util/seo";
+
+export const consultingSchema = {
```
wicksipedia: can we specify the types when we create the schemas? should help auto-completes

wicksipedia (approved):
lgtm

PR #448, opened 2023-03-24
🐛UI - Incorrect Font Showing on Booking Form

### LINKED ISSUE #427: 🐛UI - Incorrect Font Showing on Booking Form
image
**Figure: Broken fonts on booking form**

```[tasklist]
### Tasks
- [ ] Investigate if the issue occurs in any other part of the website, and then fix the broken fonts
```

Harry-Ross:
Closes #427 

URLs affected: All pages with booking form

Issue was that the booking form popup is not included in the `<Layout>` component, which applies the Open Sans font to its children. I set it so that the booking form component now also imports the Open Sans font and applies it to all of its children. 

image
**Figure: Correct fonts applied to booking form**

* Also deleted unused CSS module file


bradystroud (approved):
🛳️ :shipit: LGTM

PR #451, opened 2023-03-24
Collection - adding opportunities

### LINKED ISSUE #445: Setup collection for current employment opportunities

amankumarrr:
closes #445 

URL: `/collections/Opportunities/new`


.tina/collections/opportunities.tsx
```diff
@@ -0,0 +1,75 @@
...
+						} else {
+							duplicated = true;
+						}
+					});
+
+					if (duplicated) {
+						return "You cannot have same location multiple times!";
+					}
+				},
+			},
```
Harry-Ross: You could also write it this way - better to have `const` in the code instead of `var`.
```suggestion
				validate: (value) => {
					const locations = value?.map((x) => x.location);
	                const locArray: string[] = [];
                
	                locations?.forEach((loc) => {
		                if (!locArray.includes(loc)) {
			                locArray.push(loc);
		                } else {
			                return "You cannot have same location multiple times!";
		                }
	                });
				},
			},
```

components/util/constants.tsx
```diff
@@ -120,3 +120,56 @@ export const AustralianStatesList = [
...
+export const Job_Status = [
+	{
+		label: "Available",
+		value: "Available",
+	},
+	{
+		label: "Filled",
+		value: "Filled",
+	},
+];
```
Harry-Ross: - Move to separate file in `/components/util/constants`
- Consider changing to array of strings

Harry-Ross (changes requested):

bradystroud (approved):
LGTM💰

Harry-Ross (approved):
lgtm

PR #473, opened 2023-03-29
Booking form - adding reCaptcha validation on server side

### LINKED ISSUE #472: Validate reCaptcha in NextJs
### Description
We currently validate reCaptcha in our endpoint, but we are migrating to Power Automate's endpoint, which doesn't validate reCaptcha. Therefore, we need to validate it in NextJs' API before calling Power Automate's endpoint.

amankumarrr:
closes #472 

URL: `/consulting/software-audit` click **Book a FREE Initial Meeting**


components/bookingForm/bookingForm.tsx
```diff
@@ -95,17 +95,29 @@ export const BookingForm = ({ recaptchaKey }) => {
...
+								setContactSuccess(false);
+								router.push("/thankyou/");
+							}, 1000);
+						})
+						.catch((err) => {
+							console.error(err);
+							return alert("Failed to create lead in CRM");
+						});
+				} else {
+					return alert("Invalid reCaptcha!");
```
wicksipedia: should result in a form error not an alert dialog (bad ux)

pages/consulting/[filename].tsx
```diff
@@ -188,7 +188,8 @@ export const getStaticProps = async ({ params }) => {
 			technologyCards: technologyCardsProps,
 			marketingData: marketingSection.data,
 			env: {
-				GOOGLE_RECAPTCHA_KEY: process.env.GOOGLE_RECAPTCHA_KEY || null,
+				GOOGLE_RECAPTCHA_SITE_KEY:
+					process.env.GOOGLE_RECAPTCHA_SITE_KEY || null,
```
Comment (wicksipedia) [inline pages/consulting/[filename].tsx]:
```suggestion
					process.env.GOOGLE_RECAPTCHA_SITE_KEY || '',
```

Dockerfile
```diff
@@ -1,82 +1,88 @@
...
+
+# Build requires some environment variables -> source them as build args
+ARG GOOGLE_RECAPTCHA_KEY
+ENV GOOGLE_RECAPTCHA_KEY ${GOOGLE_RECAPTCHA_KEY}
+
+ARG GOOGLE_RECAPTCHA_SITE_KEY
+ENV GOOGLE_RECAPTCHA_SITE_KEY ${GOOGLE_RECAPTCHA_SITE_KEY}
+
+ARG CREATE_LEAD_ENDPOINT
+ENV CREATE_LEAD_ENDPOINT ${CREATE_LEAD_ENDPOINT}
```
wicksipedia: variables need to be added to `.github/workflows/template-build.yml`

.github/workflows/template-build.yml
```diff
@@ -21,6 +21,20 @@ on:
...
+      NEXT_PUBLIC_GOOGLE_ANALYTICS:
+        required: true
+      NEXT_PUBLIC_HOTJAR_ID:
+        required: true
+      NEXT_PUBLIC_HOTJAR_SV:
+        required: true
+      GOOGLE_RECAPTCHA_KEY:
+        required: true
+      GOOGLE_RECAPTCHA_KEY_v2:
+        required: true
```
wicksipedia: which one are we using?

components/bookingForm/bookingForm.tsx
```diff
@@ -22,246 +22,266 @@ import { Open_Sans } from "next/font/google";
...
-			? ACTIVE_INPUT.ClassShow
-			: ACTIVE_INPUT.None,
-		handleChange: ({ name }, e) =>
-			handleActiveInputLabel(name, e.currentTarget.value),
-	});
+    await axios
+      .post("/api/validate-token", { contactReCaptcha })
+      .then(async (res) => {
+        if (res.data.success) {
+          setInvalidReptcha("");
```
wicksipedia: this form should call the create-lead endpoint
the endpoint should be doing the recaptcha validation internally as part of input validation

.github/workflows/template-build.yml
amankumarrr: we are using v2, GOOGLE_RECAPTCHA_KEY is actually a site key, we have replaced it with GOOGLE_RECAPTCHA_SITE_KEY.

wicksipedia: if we aren't going to use it, it should be removed

.github/workflows/template-build.yml
```diff
@@ -94,6 +94,7 @@ jobs:
           NEXT_PUBLIC_TINA_BRANCH: ${{ github.head_ref || github.ref_name }}
           NEXT_PUBLIC_ZENDESK_CHAT_KEY: ${{ secrets.NEXT_PUBLIC_ZENDESK_CHAT_KEY }}
           GOOGLE_RECAPTCHA_KEY: ${{ secrets.GOOGLE_RECAPTCHA_KEY }}
+          GOOGLE_RECAPTCHA_SITE_KEY: ${{ secrets.GOOGLE_RECAPTCHA_SITE_KEY }}
```
wicksipedia: given that this is now stored in kv, can we source the secret values from there instead of using github secrets?
that way, we only have 1 set of secrets to maintain

amankumarrr: We need to log in before the build and then pass the kv secret value to it, is it a right way doing it?

amankumarrr: done

.github/workflows/template-build.yml
```diff
@@ -79,6 +79,36 @@ jobs:
...
+            --name ${{ env.KEY_VAULT }}-${{ steps.date.outputs.date }} `
+            --resource-group ${{ env.AZURE_RESOURCE_GROUP }} `
+            --template-file ./infra/keyVaultRoleAssignment.bicep `
+            --parameters `
+              "keyVaultName=${{ env.KEY_VAULT }}" `
+              "principalId=${{ secrets.AZURE_SERVICE_PRINCIPAL_OBJECT_ID }}" `
+              "principalType=ServicePrincipal" `
+              "roleName=Key Vault Secrets User" `
+            -o none
+          Write-Host '✅ SP - kv role assigned'
```
wicksipedia: this should ideally be done as part of the environment deployment

amankumarrr: I've added it in the bicep while env deployment.

wicksipedia (approved):

PR #521, opened 2023-04-19
Update - Static path on retrieving GraphQL query

### LINKED ISSUE #520: 🐛 GraphQL - Pages are not being available
Currently, Routes are not available after the first 50 pages due to pagination, we need to take all the pages into consideration while defining the dynamic path for the nextJs route.

```[tasklist]
### Tasks
- [x] Find all scenarios where this could be an issue (wherever we are generating multiple pages)
- [x] Fix all scenarios
```

amankumarrr:
closes #520 

URL: `/consulting/web-applications`

The URL above was not accessible before.

### DIFF
--- a/pages/consulting/[filename].tsx
+++ b/pages/consulting/[filename].tsx
@@ -237,9 +237,23 @@ export const getStaticProps = async ({ params }) => {
 };
 
 export const getStaticPaths = async () => {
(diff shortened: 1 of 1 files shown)


bradystroud (approved):
I dont think `aPageListData` is a great variable name, but approving to get this fix live

PR #531, opened 2023-04-21
Consulting pages - Add GPT

wicksipedia:


wicksipedia:
not progressing with this, was for demo purposes

PR #579, opened 2023-05-06
🎨Replace AsyncReturnType with InferGetStaticPropsType

### LINKED ISSUE #565: 💸Tech Debt - Replace AsyncReturnType with InferGetStaticPropsType
Right now, we have this function to get the type of the props:
image

Used like so:
image

This is copied and pasted to most of the pages to infer the type, however this isn't really an ideal solution.

NextJS has an inbuilt way of doing this - `InferGetStaticPropsType` which is used like so:
image

- [ ] Replace all instances of `AsyncReturnType` with `InferGetStaticPropsType`

KristenHu:
closed #565


bradystroud (approved):
🛳️

PR #580, opened 2023-05-06
Products page

### LINKED ISSUE #554: Pages - Products
https://www.ssw.com.au/ssw/Products/Default.aspx

Implement as a page (no need to do a collection)

KristenHu:
closed #554 

As per the PBI: #554 , we still have 2 tasks waiting for discussion:

- [ ] Legacy Products will redirect to a not-found-page, so should we keep this section or not.
image

- [x] The PBI said there's no need to do a collection, but I think creating a products index schema for this page is better.


KristenHu:
As per my conversation with @wicksipedia , task 1 for legacy products should be added as a new PBI #606  and left out of this PR currently.

pages/products/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  useEffect(() => {
+    // extract the data we need from the tina result
+    const processedData = processData(data);
+    setProductsList(processedData.productsList);
+    setSeo(processedData.seo);
+  }, [data]);
+
+  return (
+    <Layout>
+      <SEO seo={{ ...seo, canonical: "/products" }} />
```
bradystroud: Can we ge the `canonical` from the seo object?

```suggestion
       <SEO seo={seo} />
```

bradystroud (changes requested):

pages/products/index.tsx
KristenHu: > Can we get the `canonical` from the seo object?

Yes, we can, done. 😊

bradystroud (approved):
LGTM ⭐️

PR #630, opened 2023-05-12
🔊 Logging - App insight

### LINKED ISSUE #555: 🔊 Logging - App Insight in NextJs
### Description:

We need to add logs into App Insight.

There are a lot of places we could add logging, and the important one is to log leads when external endpoints could fail.

```[tasklist]
### Tasks
- [x] Update env biceps to create App Insight
- [x] Add logging to front-end
- [x] Add logging to back-end
```

amankumarrr:
closes #555


.github/.env
```diff
@@ -5,3 +5,4 @@ AZURE_RESOURCE_GROUP_LOCATION=australiaeast
 BICEP_PATH=./infra/main.bicep
 IMAGE_NAME=sswwebsite
 KEY_VAULT=kv-sswwebsite-9eb3
+NEXT_PUBLIC_APP_INSIGHT_INSTRUMENTATION_KEY=9fce8146-edbc-46d4-8542-74da9fb75503
```
wicksipedia: instrumentation keys have been deprecated - we should be using connection strings
I think it would make more sense to grab this value at build time as opposed to committing it - it is more likely to change than an resource name

.github/workflows/template-build.yml
```diff
@@ -112,6 +112,7 @@ jobs:
           GOOGLE_RECAPTCHA_KEY: ${{ secrets.GOOGLE_RECAPTCHA_KEY }}
           GOOGLE_RECAPTCHA_SITE_KEY: ${{ steps.KeyVaultSecrets.outputs.GoogleRecaptchaSiteKey }}
           KEY_VAULT: ${{ env.KEY_VAULT }}
+          NEXT_PUBLIC_APP_INSIGHT: ${{ env.NEXT_PUBLIC_APP_INSIGHT }}
```
wicksipedia: can we use clearer environment variable names... it isn't clear whether this is an resource name, endpoint, instrumentation key, or connection string

components/bookingForm/bookingForm.tsx
```diff
@@ -94,18 +96,24 @@ export const BookingForm = ({ recaptchaKey }) => {
...
+
+    const method = { Method: "Create-Lead-UI" };
     actions.setSubmitting(false);
 
     await axios
       .post("/api/create-lead", data)
       .then((response) => {
         if (response.data && !response.data.success) {
+          response.data = { ...response.data, data };
+          appInsights?.trackException(response.data, method);
```
wicksipedia:
```suggestion
          appInsights?.trackException({ input: response.data, response: data }, method);
```
the original line is difficult to read, does my suggestion convey the same intent?

infra/appService.bicep
```diff
@@ -75,6 +75,10 @@ var appSettings = [
     name: 'GOOGLE_RECAPTCHA_SITE_KEY'
     value: '@Microsoft.KeyVault(SecretUri=https://kv-sswwebsite-9eb3.vault.azure.net/secrets/Google-Recaptcha-Site-KEY)'
   }
+  {
+    name: 'APP_INSIGHT_CONNECTION_STRING'
+    value: 'InstrumentationKey=9fce8146-edbc-46d4-8542-74da9fb75503;IngestionEndpoint=https://australiaeast-1.in.applicationinsights.azure.com/;LiveEndpoint=https://australiaeast.livediagnostics.monitor.azure.com/'
```
wicksipedia: don't hardcode - the app insight resource can provide the connection string as an output

pages/_app.tsx
```diff
@@ -42,7 +43,12 @@ const App = ({ Component, pageProps }) => {
   return (
     <>
       <DefaultSeo {...NEXT_SEO_DEFAULT} />
-      <Component {...pageProps} />
+      {!isDev && (
+        <AzureAppInsights>
+          <Component {...pageProps} />
+        </AzureAppInsights>
+      )}
+      {isDev && <Component {...pageProps} />}
```
wicksipedia: I would've expected that the container would always be setup and app insights enabled/disabled internally as required

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.22.19",
   "private": true,
   "scripts": {
-    "dev": "tinacms dev -c \"next dev\"",
+    "dev": "tinacms dev -c \"node --require .\\appInsight-api.js node_modules/next/dist/bin/next dev\"",
```
wicksipedia: would not have expected that this would be required

package.json
```diff
@@ -37,9 +37,12 @@
     "@fortawesome/react-fontawesome": "^0.2.0",
     "@headlessui/react": "1.7.14",
     "@headlessui/tailwindcss": "^0.1.3",
+    "@microsoft/applicationinsights-react-js": "^3.4.2",
+    "@microsoft/applicationinsights-web": "^2.8.5",
     "@tailwindcss/typography": "^0.5.9",
     "animate-css-grid": "^1.5.1",
     "aos": "^2.3.4",
+    "applicationinsights": "^2.6.0",
```
wicksipedia, line 45:
```suggestion
```

appInsight-api.js
```diff
@@ -0,0 +1,15 @@
+if (typeof process.env.APP_INSIGHT_CONNECTION_STRING !== "undefined") {
```
wicksipedia: the setup of this file isn't in the app insight docs, is it really required?

services/appInsight_ui.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+let browserHistory = null;
+if (typeof document !== "undefined") {
+  browserHistory = createBrowserHistory();
+}
+const reactPlugin = new ReactPlugin();
+
+const appInsights = new ApplicationInsights({
+  config: {
+    instrumentationKey: process.env.NEXT_PUBLIC_APP_INSIGHT_INSTRUMENTATION_KEY,
```
wicksipedia: according to the docs, we can use connection string here

wicksipedia (commented):
Have a look at https://learn.microsoft.com/en-us/azure/azure-monitor/app/javascript-framework-extensions?tabs=react - the method described in the docs is less invasive

services/appInsight_ui.tsx
amankumarrr: fixed it

package.json
amankumarrr: this appInsight is for the backend. We load it at compile time.

package.json
amankumarrr: In order to use appInsight with Next API, it is necessary to load it at compile time. There is no other way to load appInsight and run Next together. This means that appInsight must be loaded while running the Next server.

infra/appService.bicep
amankumarrr: you're right!, I've fixed it with it

components/bookingForm/bookingForm.tsx
amankumarrr: It only accepts IExceptionTelemetry or Axios response. So your suggestion above would not work.

pages/_app.tsx
amankumarrr: done - so instead of checking it here we are loading the appInsight based on the env variable

appInsight-api.js
amankumarrr: It is because we have to provide that variable even before it loads .env file from the Next Server. So for local, we have to create .env file on the root folder to access the variable otherwise it will fail on both local and on the PR slot if that variable doesn't exist. We can also turn on the logging for testing purposes by adding the key in the app config and restarting the App.

.github/workflows/template-build.yml
```diff
@@ -112,6 +114,7 @@ jobs:
           GOOGLE_RECAPTCHA_KEY: ${{ secrets.GOOGLE_RECAPTCHA_KEY }}
           GOOGLE_RECAPTCHA_SITE_KEY: ${{ steps.KeyVaultSecrets.outputs.GoogleRecaptchaSiteKey }}
           KEY_VAULT: ${{ env.KEY_VAULT }}
+          NEXT_PUBLIC_APP_INSIGHT_CONNECTION_STRING: ${{ secrets.NEXT_PUBLIC_APP_INSIGHT_CONNECTION_STRING }}
```
wicksipedia: can we get the connection string via the cli instead of saving it as a secret?

components/bookingForm/bookingForm.tsx
wicksipedia: I think the multiple reference to "data" is throwing me off. Can we then rename the variable or make it more clear via a type?

infra/create-pr-environment.bicep
```diff
@@ -26,6 +26,13 @@ module appServiceSlot 'appSerivce-create-slot.bicep' = {
   }
 }
 
+module appInsight 'appInsight.bicep' = {
```
wicksipedia: can we have a non-prod app insight instance and wire it up like prod

package.json
wicksipedia:
```suggestion
    "dev": "tinacms dev -c \"node --require ./appInsight-api.js node_modules/next/dist/bin/next dev\"",
```

services/appInsight_ui.tsx
```diff
@@ -0,0 +1,45 @@
...
+  browserHistory = createBrowserHistory();
+}
+
+const isDev = process.env.NODE_ENV === "development";
+
+const reactPlugin = new ReactPlugin();
+
+const appInsights = new ApplicationInsights({
+  config: {
+    instrumentationKey: process.env.NEXT_PUBLIC_APP_INSIGHT_CONNECTION_STRING,
```
wicksipedia: connection string is being set to the instrumentation key

services/appInsight_ui.tsx
```diff
@@ -0,0 +1,45 @@
...
+    extensionConfig: {
+      [reactPlugin.identifier]: { history: browserHistory },
+    },
+  },
+});
+
+const AzureAppInsights = ({ children }) => {
+  useEffect(() => {
+    if(!isDev)
+    appInsights.loadAppInsights();
```
wicksipedia:
```suggestion
    if (!isDev) {
      appInsights.loadAppInsights();
    }
```

services/appInsight_ui.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { useEffect } from "react";
```
wicksipedia: I think the convention for this file would be to put it in `/context/appInsights.ts`

services/appInsight_ui.tsx
```diff
@@ -0,0 +1,45 @@
...
+  AppInsightsContext,
+} from "@microsoft/applicationinsights-react-js";
+import { createBrowserHistory } from "history";
+
+let browserHistory = null;
+if (typeof document !== "undefined") {
+  browserHistory = createBrowserHistory();
+}
+
+const isDev = process.env.NODE_ENV === "development";
```
wicksipedia: I think we should load up AppInsights when a connection string is provided.

If one is not provided, log a message to console

appInsight-api.js
wicksipedia:
```suggestion
if (!!process.env.APP_INSIGHT_CONNECTION_STRING) {
```

this should check to see if the connection string is present and has value

context/appInsight_ui.tsx
wicksipedia, line 1: can we rename this file, there is no UI associated with it

- client would be a better suffix
- file name should be kebab case

wicksipedia (changes requested):

wicksipedia (approved):

PR #665, opened 2023-05-22
Standardize-domain-name-for-seo

wicksipedia:
- front door redirects to www so we want to be consistent

Relates to SSWConsulting/SSW.Website#1205

Affected routes:

- N.A


bradystroud (approved):
🚢

PR #742, opened 2023-06-05
Rename repo - drop the v3

### LINKED ISSUE #750: Rename repo - drop the v3
### Description:
We have renamed the old repo with zz as a prefix, now we need to drop the v3 for our new repo.

wicksipedia:
Renaming repo as there should be only 1 website

Fixes #750


bradystroud (approved):
lgtm 🚀

PR #758, opened 2023-06-06
💄 Add icon to Google Map link

### LINKED ISSUE #757: 💄 Google Maps icon
CC: @adamcogan @bradystroud @wicksipedia @tiagov8 

As per my conversation with @adamcogan , it appears better to have an icon placed to the left of the underline link. https://www.ssw.com.au/offices/newcastle

- [x] Add Googles Maps icon for the office page

image

Aibono1225:
Fixes #757 

Changes I made
- Added google maps directions to Hangzhou and Strasbourg offices
- Added an icon on the left of the google map link

Affected routes:
- /offices/brisbane
- /offices/melbourne
- /offices/newcastle
- /offices/strasbourg
- /offices/sydney
- /offices/hangzhou

image


content/offices/brisbane.mdx
```diff
@@ -57,7 +57,8 @@ aboutUs: >
   Server or Visual Studio Online, and Microsoft Azure.
 map: /images/offices/brisbane/SSW-brisbane-map-wide.jpg
 directions: >
-  #### [Level 1, 471 Adelaide St, Brisbane, QLD 4000,
+  ####  [<CustomImage altText="map-icon" height={16} width={16}
```
wicksipedia: given that all this is repeated for each use of the directions prop - I think this should be part of the page template instead of repeated for each city

styles.css
```diff
@@ -3,92 +3,96 @@
...
+    padding: 8px;
+  }
+
+  .customTable li {
+    list-style-type: square;
+  }
+
+  img[alt="map-icon"] {
+    margin: 0 6px 4px 0;
+  }
```
wicksipedia: We are trying to use utility css where possible. When you move the image to the page template - can you use the utility classes that replicate this margin setting

wicksipedia (changes requested):

content/offices/brisbane.mdx
Aibono1225: I tried to add the SVG image to the page template, but it doesn't work. This is because we are using TinaMackdown to render the entire directions section and cannot insert the img tag inside the h4 tag.

styles.css
Aibono1225: Given that we're unable to move the image to the page template, I think it would be better to maintain this CSS for the map icon🤔

wicksipedia (approved):
If you are happy with my changes, then it is all approved and good to merge 👍

Issue #768, opened 2023-06-07
Add China website link to Hangzhou office page


hveraus:
URL: https://www.ssw.com.au/offices/hangzhou

1. Like on SSW France’s office page, we need an URL to https://www.ssw.com.cn/?utm_source=referral&utm_medium=sswaustralia. So it’s consistent with https://www.ssw.com.au/offices/strasbourg#Directions

PR #769, opened 2023-06-07
📝🔀 Make PR lint check skip if only content is changed

### LINKED ISSUE #777: 🐛 Skip PR check if content PR
This is still a required check, so it needs to run on every PR.

Fixes the problem on https://github.com/SSWConsulting/SSW.Website/pull/872

bradystroud:
This is still a required check, so it needs to run on every PR.
Not sure if there is a better way to do this.

We could alternatively make this a not required check

Fixes the problem on #768

### DIFF
--- a/.github/workflows/pr-lint-pr.yml
+++ b/.github/workflows/pr-lint-pr.yml
@@ -21,20 +21,20 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v2
 
-      - name: Check for non-content/public changes
+      - name: Check paths
         id: changes
         uses: dorny/paths-filter@v2
         with:
           filters: |
-            nonContentPublic:
-              - '!content/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-lint-pr.yml
```diff
@@ -19,12 +16,31 @@ jobs:
...
+          should_skip=true
+
+          for file in $modified_files; do
+            if [[ $file != content/* && $file != public/* ]]; then
+              should_skip=false
+              break
+            fi
+          done
+
+          echo "should_skip=$should_skip" >> $GITHUB_ENV
```
wicksipedia: can we use an action like https://github.com/marketplace/actions/has-changed-path to do the check instead?

bradystroud: Done

PR #799, opened 2023-06-13
Consulting enquiry - Disable Button and spinner on loading

### LINKED ISSUE #728: ✨ - Consulting - Disable button on enquiry submission
### Description:
On submission, the user can hit the button multiple times on submit button which could cause the duplication of leads.

```[tasklist]
### Tasks
- [x] Disable the button until it gets the response/error
- [x] Show spinner or loading sign (get it approved first)
```
Enquiry Form

**Figure: Enquiry button that needs to be disabled upon submission**

amankumarrr:
fixes #728 

#### Description:
Button will be disabled upon submitting and spinner will be replaced with the current icon to show if it's loading with the text PROCESSING.
- Button will be disabled until it gets any response/error
- Cursor will be changed to CROSS and text SUBMIT to PROCESSING

#### Screenshot:
NewSpinner
**GIF: Showing spinner on loading**


components/spinner/spinner.tsx
bradystroud, line 1: This is not how we do icons. Please use FontAwesome.

bradystroud (changes requested):

components/spinner/spinner.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export const Spinner = () => (
+  <svg
```
jaydenalchin: From a UX perspective this is looking great! But we should use fontawesome icons instead of svgs 

e.g. class="fa-solid fa-spinner"

jaydenalchin (approved):

components/spinner/spinner.tsx
amankumarrr: done - I'm using Font Awesome Icon now, GIF is updated in the description above.

styles.css
```diff
@@ -67,13 +67,16 @@
...
+    @apply items-center justify-center rounded border-none bg-sswRed px-4 py-3 align-middle text-sm font-normal uppercase leading-[18px] text-white no-underline !outline-sswRed hover:!bg-[#8e2c2c] hover:!text-white;
   }
 
   .m-icon {
     @apply mr-2;
   }
 
+  .sswSpinner {
+    @apply mr-2 h-4 w-4 animate-spin fill-sswRed text-white dark:text-gray-600;
+  }
```
bradystroud: 💣

```suggestion
```

bradystroud (approved):
LGTM 🛫

PR #801, opened 2023-06-13
💸Tech Debt - Save Times Too Long

### LINKED ISSUE #798: 💸 Tech Debt - Save Times Too Long
**Description**  
At the moment, upon save, Prettier takes ~10 to 15 seconds per save, which has been quite troublesome, with the temporary solution for many on the team to turn off format on save. This is troublesome, as it means that the Prettier and ESLint style rules are not enforced on save, and thus decreasing code quality.

**Proposed solution**  
This can be fixed by installing [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) as per https://prettier.io/docs/en/integrating-with-linters.html.

Harry-Ross:
There have been significant increases in file save times, causing devs to either disable formatting on save or having to restart their VSCode. This isn't ideal, and leads to a much worse developer experience. 

Fixes #798 

Affected routes: N/A
**NOTE:** Upon merging, we must notify developers of change in workflow


bradystroud (approved):
lgtm

PR #807, opened 2023-06-14
PR Push Deploy - Adding delay before page insight

### LINKED ISSUE #785: 💚 PR check - Pagespeed insights fails on first run
The PageSpeed insights PR check fails on its first run.

I think this is because its needs the PR environment to be deployed, and there is a short period after the deploy has completed but 
the site is not actually live.

```[tasklist]
### Tasks
- [x] Investigate
- [x] Add a delay before running PageSpeed Insights
- [x] Follow-up with this PR https://github.com/wicksipedia/pagespeed-insights/pull/41, test it when it gets merged
```

**Figure: PageSpeed Insights failing**

amankumarrr:
Fixes #785

### DIFF
--- a/.github/workflows/pr-push-deploy.yml
+++ b/.github/workflows/pr-push-deploy.yml
@@ -78,17 +78,27 @@ jobs:
         run: |
           $slotURL = "${{ needs.pr-deploy.outputs.url }}"
           $delaySeconds = 30
+          $maxWaitSecs = 300
+          $totalDelay = 0
 
           while ($true) {
+            try {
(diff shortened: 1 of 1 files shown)


bradystroud (changes requested):
Hey @amankumarrr, well done on the fix! 🎉

Here are a couple of things I noticed:

1. The delay part in the code could use a comment. It'll help others understand why it's there. 🤔👀
2. A bit more description in the PR would make it easier to review. 👁️📝

Cheers! 👍🙂

PR #821, opened 2023-06-16
✨Newsletters - File Upload on TinaCMS

### LINKED ISSUE #626: Newsletters - Upload file feature on TinaCMS
We would like Penny to have as little action as possible to upload the new newsletters.
Currently, she has to enter newsletter information in TinaCMS (file name, month...), but she also has to manually add the HTML file to the public/newsletters/[year] folder.

Using TinaCMS's Media Manager does not meet our needs as any uploaded file will always be placed in the root of the /public folder.
An issue is currently open on the TinaCMS repository to address this : https://github.com/tinacms/tinacms/issues/3813

In the meantime, until we can utilize an internal feature within TinaCMS, we need to find another way to make Penny's work easier.  

1. go to http://localhost:3000/admin/index.html#/collections/newsletters | any json file
2. create a upload feature that would put the selected file on the right spot : public/newsletters/[the years related to json file name]
3. listing the file currently on the folder
4. let user delete a file if necessary

Harry-Ross:
At the moment file upload can't be achieved through the Tina UI, the workflow would involve first commiting the desired HTML and image files in a branch, then updating the link in the Tina UI. This is not ideal as there is no hard link between the files and the TinaCMS content. 

Originally, I developed a custom component to upload files, however the `type: "image"` component allows file upload of not just images, but HTML files too, so I made the simple addition of changing the type of the URL in the TinaCMS content from `string` to `image`, which are functionally the same - as `image` is stored as a string in the TinaCMS content. 

Fixes #626 

Affected routes:

- `/newsletters`

image
**Figure: Added "HTML File" and "Images (optional) fields to the newsletter object**


.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
-	"editor.formatOnSave": true,
+	"editor.formatOnSave": false,
```
bradystroud: Change this back

components/admin/customFileUpload.tsx
bradystroud, line 1: Is this used?

.tina/config.tsx
```diff
@@ -35,7 +35,7 @@ const config = defineStaticConfig({
     // this is the config for the tina cloud media store
     tina: {
       publicFolder: "public",
-      mediaRoot: "images",
+      mediaRoot: "",
```
bradystroud: Will this affect anything else? (i think it will)

bradystroud (changes requested):
Nice work 💪 its good that the image component worked for any file type, but this seems like something Tina should fix (rename it to `file`)
Have you made a feature request?

.tina/config.tsx
bradystroud: Yep, it broke all images as expected.

see https://app-sswwebsite-9eb3-pr-821.azurewebsites.net/

Figure: All images on the home page are broken 😕
**Figure: All images on the home page are broken 😕**

bradystroud (approved):
lgtm 🛶

PR #868, opened 2023-06-26
Azure Resouces - adding tags

### LINKED ISSUE #748: ♻️ Infra - Add Cost-Category tag to all resources
**Describe the bug**
Not exactly a bug, but all resources used in SSW need to be tagged with a cost-category of 'dev/test', 'value' or 'core'
dev/test - for things you are working on
value - for things that aren't required but are important
core - for things that are core to the business.  I.E. website

As per https://ssw.com.au/rules/azure-budgets/

1. Please update the infra deployment for this so that my Azure Cost Analysis weekly journey is easy 😊
2. Get 'Checked by Warwick'

amankumarrr:
Fixes #748 

#### Description:
Implemented a tagging for Azure resources, utilizing 'dev/test,' 'core,' and 'value' tags to effectively categorize and track their associated costs.

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -30,9 +30,12 @@ param dockerImage string = 'sswwebsite'
 
 param now string = utcNow('yyyy-MM-ddTHH-mm')
 
-var tags = {
-  // change this to 'core' when you are ready to deploy to production
-  'cost-category': 'dev/test'
+var core = {
+  'cost-category': 'core'
+}
+
+var value = {
+  'cost-category': 'value'
 }
 
 module acr 'acr.bicep' = {
@@ -40,7 +43,7 @@ module acr 'acr.bicep' = {
   params: {
     projectName: projectName
     location: location
-    tags: tags
+    tags: core
   }
 }
 module keyVault 'keyVault.bicep' = {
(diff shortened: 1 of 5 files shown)


leahy268 (approved):
Looks ok to me although I'm not completely familiar with the Infra in the repo

infra/main.bicep
```diff
@@ -30,31 +30,36 @@ param dockerImage string = 'sswwebsite'
...
+  'cost-category': 'value'
 }
 
 module acr 'acr.bicep' = {
   name: 'acr-${now}'
   params: {
     projectName: projectName
     location: location
-    tags: tags
+    tags: core
```
wicksipedia: if this is always `core` - then it should be set in the module and not determined outside in the `main` file

infra/main.bicep
```diff
@@ -30,31 +30,36 @@ param dockerImage string = 'sswwebsite'
...
   }
 }
 module keyVault 'keyVault.bicep' = {
   name:'keyVault-${now}'
   params: {
     projectName: projectName
     location: location
+    tags:core
```
wicksipedia: if this is always `core` - then it should be set in the module and not determined outside in the `main` file

wicksipedia (approved):
Approved with suggestions

infra/main.bicep
amankumarrr: Done - yes it makes sense

infra/main.bicep
amankumarrr: Done - yes it makes sense

wicksipedia (approved):

Issue #248, opened 2023-02-21
⚡Google Tag Manager - review tags being used


wicksipedia:
Liase with @camillars

GTM currently loads a lot of tags - do an audit and clean up where possible

Install MS Clairty at the same time #722

amankumarrr:
@bradystroud 

We are using following tags with our GTM:

![Image](https://github.com/SSWConsulting/SSW.Website/assets/71385247/9d5968bc-ce17-4b9c-a430-ee6bfae88d22)

**Figure: list of tags with GTM** - https://tagmanager.google.com/?authuser=1#/container/accounts/362468/containers/407674/workspaces/1000037/tags

@camillars, we need a list of tags that are redundant, need to clean up to improve the performance of the website. 
@wicksipedia this #722 issue has been closed as the Microsoft Clarity is already used as a tag in GTM.

camillars:
Hi @amankumarrr 

here's a preview list. Yellow means I am unsure (under investigation), green stays, and red can go.

![image](https://github.com/SSWConsulting/SSW.Website/assets/53064613/b1679c3b-ca57-42e9-bff3-72a5fdf20ffc)

amankumarrr:
@bradystroud @wicksipedia 

Hi @camillars 

Here is the preview of tags after removing the mentioned ones above. Please let me know if we need some more cleanup.

<img width="947" alt="image" src="https://github.com/SSWConsulting/SSW.Website/assets/71385247/c193b703-aea1-45df-a9cd-17e5e5f4a8dc">

**Figure: GTM tags after a clean-up**

wicksipedia:
Is AppInsights required? At minimum, it looks like it is misconfigured ATM

<img width="1436" alt="CleanShot 2023-06-27 at 10 34 32@2x" src="https://github.com/SSWConsulting/SSW.Website/assets/600044/d04fad7f-fce9-4675-93fc-d01f1cd47aab">

wicksipedia:
should GTM be loaded on `lazyOnload` instead of `afterInteractive` (default)?
see: https://nextjs.org/docs/pages/api-reference/components/script#strategy

it looks like it is the biggest contributor to 3rd patry code impact
<img width="1317" alt="CleanShot 2023-06-27 at 10 37 20@2x" src="https://github.com/SSWConsulting/SSW.Website/assets/600044/f6b60e7b-ba38-40a4-bcec-37c0cea3fc30">

amankumarrr:
[...]

PR #875, opened 2023-06-27
GTM - Lazy loading for scripts

amankumarrr:
Related to #248 

**Description:**
By default, it was loading GTM scripts as afterInteractive strategy, which consumes time while loading the page, now we are loading the scripts as LazyLoading strategy as it will be injected into the HTML client-side during browser idle time.

Affected routes:

- `/`


amankumarrr:
closed this PR as Lazy loading didn't improve the performance

PR #917, opened 2023-07-03
PR - Main Infra check

### LINKED ISSUE #897: 🐛 PR - Fix check to ensure main infra code works
**Describe the bug**
There is a failing check https://github.com/SSWConsulting/SSW.Website/actions/workflows/pr-infra.yml

```[tasklist]
### Tasks
- [x] Fix the title
- [x] Fix the triggers (change to PR)
- [x] Add the missing parameters
```

amankumarrr:
Fixes #897 

**Description:**
Removing the duplicate name of workflows, added testing for the main's BICEP with the pull request, and fixing the Service Principal ID error with the flow.


.github/workflows/pr-main-infra-check.yml
```diff
@@ -1,7 +1,7 @@
-name: Main - Infra - Build and deploy
+name: Main - Infra Check - Build and deploy
```
bradystroud:
```suggestion
name: PR - Infra Check - Build and deploy
```

bradystroud (changes requested):

.github/workflows/pr-main-infra-check.yml
amankumarrr: But we are checking main's infra there, do you think we should make these changes?

bradystroud: Looking at the other workflows, I think the naming scheme is {{ TRIGGER }} - {{ ACTION }}

Maybe a better name would be `PR - Test main infra`

amankumarrr: Yes it makes more sense.

bradystroud (approved):
lgtm

wicksipedia (approved):

PR #926, opened 2023-07-04
🐛 Footer - DayJS RelativeTime Error

### LINKED ISSUE #899: 🐛 Consulting index error
**Describe the bug**
When you load the consulting index page, there is an error.

**To Reproduce**
1. Go to `/consulting`

image

Harry-Ross:
* Added global DayJS plugin additions to ensure bug cannot occur again

From my investigation, the reason why the `/consulting` route had a bug with `.fromNow()` function from the `dayjs/plugin/relativeTime` module, but not the other pages was because the `UpcomingEvents` component was imported by using the `<Blocks>` component, which contains `dayjs.extend(relativeTime)`. This wasn't noticed as most of the pages have the `<Blocks>` component in the page, which therefore loaded `UpcomingEvents` and the DayJS module extension. I have fixed this by having a single source of truth in the root app file for which DayJS modules have been loaded, which should make it clear to both devs and the TypeScript compiler which modules are available for use in the project.

I found the desired configuration at https://github.com/iamkun/dayjs/issues/1577. This means that by using `dayjs.extend()`, each subsequent use of DayJS will have the plugin used with the function.

Fixes #899

Affected routes:

- All (`<Footer />`)

### DIFF
--- a/components/blocks/upcomingEvents.tsx
+++ b/components/blocks/upcomingEvents.tsx
@@ -1,7 +1,5 @@
 import dayjs from "dayjs";
-import isBetween from "dayjs/plugin/isBetween";
-import relativeTime from "dayjs/plugin/relativeTime";
-import utc from "dayjs/plugin/utc";
+// import relativeTime from "dayjs/plugin/relativeTime";
 import Image from "next/image";
 import Link from "next/link";
 import { useEffect, useState } from "react";
@@ -10,10 +8,6 @@ import { tinaField } from "tinacms/dist/react";
 
 import axios from "axios";
 import { EventInfo, LiveStreamBannerInfo } from "../../services/server/events";
-
-dayjs.extend(utc);
-dayjs.extend(isBetween);
-dayjs.extend(relativeTime);
 
 export const UpcomingEvents = ({ data }) => {
   const [events, setEvents] = useState([]);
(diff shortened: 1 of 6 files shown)


components/blocks/upcomingEvents.tsx
```diff
@@ -1,7 +1,5 @@
 import dayjs from "dayjs";
-import isBetween from "dayjs/plugin/isBetween";
-import relativeTime from "dayjs/plugin/relativeTime";
-import utc from "dayjs/plugin/utc";
+// import relativeTime from "dayjs/plugin/relativeTime";
```
bradystroud:
```suggestion
```

bradystroud (approved):
Thanks @Harry-Ross 

> From my investigation, the reason why the /consulting route had a bug with .fromNow() function from the dayjs/plugin/relativeTime module, but not the other pages was because the UpcomingEvents component was imported by using the <Blocks> component, which contains dayjs.extend(relativeTime). This wasn't noticed as most of the pages have the <Blocks> component in the page, which therefore loaded UpcomingEvents and the DayJS module extension. I have fixed this by having a single source of truth in the root app file for which DayJS modules have been loaded, which should make it clear to both devs and the TypeScript compiler which modules are available for use in the project.
> 
> I found the desired configuration at https://github.com/iamkun/dayjs/issues/1577. This means that by using dayjs.extend(), each subsequent use of DayJS will have the plugin used with the function.

When you close the issue, can you copy this info into it?

PR #944, opened 2023-07-05
📲 Products - SSW Rewards Page

### LINKED ISSUE #890: ✨ Migrate Rewards page
https://www.ssw.com.au/ssw/Rewards/

The route should be `/products/ssw-rewards`

Harry-Ross:
* Added the new SSW Rewards page from https://www.ssw.com.au/ssw/rewards/
* Added "TweetEmbed" component + new API route, as the Twitter oEmbed API does not support CORS, which means there needs to be an intermediary API route, instead of requesting it straight from the browser as per: https://stackoverflow.com/questions/35879943/twitter-api-authorization-fails-cors-preflight-in-browser

Fixes #890

Affected routes:

- `/products/rewards`

image


bradystroud:
image

bradystroud (approved):
L G T M

PR #1031, opened 2023-07-14
Add editable tagline to testimonial component

### LINKED ISSUE #604: What do people say | missing a sentence under the title
~~@BrookJeynes please give this a decription~~ - Done

As per the figma design, the "What do people say" section should have some text below the header.

image
**Figure: Design from Figma board**

jackreimers:
Adds an editable tagline to the testimonial row component.
The tagline can be edited per page in Tina.

Fixes #604 

Affected routes:

- `/consulting/*`
- `/training/internship-fullstack`

image


Harry-Ross (changes requested):
Hi Jack, I think instead of having a schema for the testimonial row component, it would be better to add a `rich-text` field to the consulting and training pages, that is rendered above the testimonial row like so:
image

This is because the testimonial fetching is hard coded into each page, and is loaded through getStaticProps. This is also better because it means that non-tech people can edit the "What do people say?" section.

jackreimers:
@Harry-Ross I don't see a situation where the title on the testimonial would need to be changed, but the tagline does need to be different on each page as Uly doesn't want a tagline on the consulting pages.

Having the title and tagline in a markdown would mean we would have to set the title for each page as well which could be time consuming.

@bradystroud

Harry-Ross:
> @Harry-Ross I don't see a situation where the title on the testimonial would need to be changed, but the tagline does need to be different on each page as Uly doesn't want a tagline on the consulting pages.
> 
> Having the title and tagline in a markdown would mean we would have to set the title for each page as well which could be time consuming.
> 
> @bradystroud

Upon reflection, you are right - approved

Harry-Ross (approved):

PR #1049, opened 2023-07-17
🐛 1045 - Add default width and height values to custom image

### LINKED ISSUE #1045: 🐛 Exception when adding image to Custom Image component
**Describe the bug**
When embedding a Custom Image component, if the image is added before entering a value for the width then an exception occurs.

**To Reproduce**
1. Create a new Custom Image component in a rich text field.
2. Edit the component and add an image.
3. Exception occurs because the width field has no value.

**Expected behavior**
The fields of the component should be able to be filled in any order without an exception.

**Screenshots/Video**
image

jackreimers:
Adds a default width and height value to the Image in the Custom Image component to prevent an exception when there is no value set.

Related to #1045

### DIFF
--- a/components/blocks/customImage.tsx
+++ b/components/blocks/customImage.tsx
@@ -7,8 +7,8 @@ export const CustomImage = ({ data }) => {
       <Image
         src={data.src}
         alt={data.altText}
-        height={data.height}
-        width={data.width}
+        height={data.height ?? 400}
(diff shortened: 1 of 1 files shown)


bradystroud (approved):
As per our conversation, this will always get overridden anyway beacuse it is a required field in TinaCMS.
It is just here for when the user adds the image before they add the height and width

PR #1136, opened 2023-07-28
🐛 Canonical - Added Canonical URLs to Routes 

### LINKED ISSUE #1123: 🐛 SEO - Canonical URLs not being set properly
**Describe the bug**
As per my conversation with @bradystroud in [this PR](https://github.com/SSWConsulting/SSW.Website/pull/1119#discussion_r1274461455), currently most of the templates are overriding the canonical URL in the SEO schema. 
This is bad if we manually set a canonical URL because it will be overwritten.
Some templates are also not setting a canonical URL at all, for example `/training`

```[tasklist]
### Tasks
- [x] Ensure templates only set a canonical URL if one is not manually set
- [x] Ensure all templates are setting a canonical URL
```

**Expected behavior**
Every page should have a canonical URL and no manually entered one should be overriden.

Harry-Ross:
* Added canonical URLs to sites that don't have them

Fixes #1123 

Affected routes:

- `/industry/*`
- `/events/*`
- `/offices/*`
- `/products/*`
- `/training/*`

### DIFF
--- a/pages/events/[filename].tsx
+++ b/pages/events/[filename].tsx
@@ -129,9 +129,8 @@ export const getStaticProps = async ({ params }) => {
     (t) => t.node
   );
 
-  const canonical = `${tinaProps.data.global.header.url}events/${params.filename}`;
-  if (tinaProps.data.events.seo) {
-    tinaProps.data.events.seo.canonical = canonical;
+  if (tinaProps.data.events.seo && !tinaProps.data.events.seo.canonical) {
+    tinaProps.data.events.seo.canonical = `${tinaProps.data.global.header.url}events/${params.filename}`;
   }
 
   return {
--- a/pages/events/[filename].tsx
+++ b/pages/events/[filename].tsx
@@ -129,6 +129,11 @@ export const getStaticProps = async ({ params }) => {
     (t) => t.node
   );
 
+  const canonical = `${tinaProps.data.global.header.url}events/${params.filename}`;
+  if (tinaProps.data.events.seo) {
+    tinaProps.data.events.seo.canonical = canonical;
+  }
+
   return {
     props: {
       data: tinaProps.data,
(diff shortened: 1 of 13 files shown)


pages/events/[filename].tsx
```diff
@@ -129,6 +129,11 @@ export const getStaticProps = async ({ params }) => {
     (t) => t.node
   );
 
+  const canonical = `${tinaProps.data.global.header.url}events/${params.filename}`;
```
Comment (jackreimers) [inline pages/events/[filename].tsx]: Is this used elsewhere? Otherwise I'd define it inside the if statement

jackreimers (changes requested):

pages/events/[filename].tsx
```diff
@@ -129,6 +129,11 @@ export const getStaticProps = async ({ params }) => {
...
+  if (tinaProps.data.events.seo) {
```
Comment (jackreimers) [inline pages/events/[filename].tsx]: This doesn't check if a URL has been set in Tina already and will always overwrite it
I'd do something like:
if (seo && (seo?.canonical === null || seo?.canonical === "")) {

jackreimers (changes requested):

pages/offices/index.tsx
```diff
@@ -141,6 +141,13 @@ export const getStaticProps = async () => {
   const testimonial =
     testimonials[Math.floor(Math.random() * testimonials.length)].Testimonial;
 
+  if (
+    tinaProps.data.officeIndex.seo &&
+    !tinaProps.data.officeIndex.seo.canonical
+  ) {
+    tinaProps.data.officeIndex.seo.canonical = "/offices";
```
jackreimers: I think these need to be fully qualified in case there are differences in the base URL, eg http vs https
Same for consulting and products

jackreimers (changes requested):

pages/offices/index.tsx
Harry-Ross: Done, had to add extra query for consulting and products as they current operate by making a `connection` query

jackreimers (approved):
LGTM

PR #1183, opened 2023-08-07
PR - Adding new check for checking PR and slots counts

### LINKED ISSUE #1175: 🐛 WebApp Slots - Old PR slots are still running
**Describe the bug**
We currently have 3 active PRs, but from the resources, it seems like 9 slots that have been running, which implies that some pipelines are not working as expected or need to be cleaned up some old slots to avoid extra cost. 

**Expected behavior**
For every active PR, there should be a slot that gets deleted when it gets merged.  

image

**Figure: SSW website's 9 slots that are running** 

```[tasklist]
### Tasks
- [x] Investigate the issue
- [x] Fix the issue
- [x] Clean up the slots and associated roles
```

amankumarrr:
Related: #1175 

- Modified the `pr-delete-close-env` to make it a callable flow 
- Added new flow to clean up the slots every Monday day at 12 noon 

#### Check Active PRs and Slots - Flow 

1. It fetches all the slots that are running under our AppService
2. It fetches all the Active PRs from our repo 
3. It checks if we have any slot that is running and has not any active Pull Request
4. Upon finding a slot that is running without a Pull Request, it calls the second flow `pr-delete-close-env` with an input of slot ids that need to be deleted
5. If in step 3, it finds no slot running without a PR, it will skip the 4th step and terminate the flow

#### Impacted Flows:
`pr-close-delete-env`

This flow is now capable of handling the following triggers:

1. Workflow_Dispatch
2. Call from another flow
3. On PR closed 
 
All triggers can be handled either by giving single id or ids separated by space 

#### Tested Scenarios:

✅ Manually triggered `pr-close-delete-env` flow for (Workflow_Dispatch)
✅ Rerun the flow after deleting the slots
✅ Ran the flow with multiple and single slot ids
 
image

**Figure: Deleted 2 slots in parallel**


amankumarrr:
closing this PR because I deleted the ACR Role with the flow and now it's not allowing me to update or recreate the flow.

PR #1197, opened 2023-08-09
Testpr close issue

amankumarrr:


amankumarrr:
closing it to test the scenario when it fails to delete the slot

PR #1217, opened 2023-08-11
New Page - Events Index

### LINKED ISSUE #1056: Index page for /events
As per my conversation with @camillars this page should allow users to link to certain events. These links should scroll to the event.

This page will need to fetch data from the SharePoint events (e.g. User Groups), and the training events in Tina (e.g. Superpowers)

```[tasklist]
### Tasks
- [x] Migrate the page https://www.ssw.com.au/ssw/Events/?tech=all&type=all
- [x] Keep the same filters avaliable
- [x] Each event should also be an anchor link
```

Harry-Ross:
* Adds new `/events` index page
* Abstracted the filter component on `/employment` for use on `/events`
* Added single document collection functionality to index pages (https://tina.io/docs/editing/single-document-collections/)

Fixes #1056

Affected routes:

- `/events`
- `/employment`

**TODO:** Add done video, screenshots

**Questions:**
1. Should past events be loaded on page load or conditionally loaded?
2. Do we need to maintain state of filters when switching between tabs?


bradystroud (approved):

bradystroud:
> 1. Should past events be loaded on page load or conditionally loaded?

Ideally only when you navigate to that tab, but this is good enough for now, improvements can be made later


> 2. Do we need to maintain state of filters when switching between tabs?

Again, this would be nice but not MVP

PR #1265, opened 2023-08-18
💄 UX - Table Layout Improvements

### LINKED ISSUE #1264: 💄 UX - Table Layout Improvements
As per my conversation/email with @pierssinclairssw a couple of weeks ago, there should be an addition of the table header added to to cells somehow, both in the list of cells, and in the cells themselves.

```[tasklist]
### Tasks
- [x] Add this functionality
- [x] Test please from Piers
```

Harry-Ross:
* Added label when editing cell to denote which table heading you are currently editing
* Refactored value discovery function to allow for reuse
* Removed "source.organiseImports" from settings, replaced with "source.sortImports" - this is because it was erraneously removing an "unused import" when in reality it was being used. This means that unused imports now have to manually removed, but I believe this will improve DX, as what usually happens is you import some code and it is removed before you can use it when you press save

Fixes #1264 

Affected routes:

- `/admin`

image


bradystroud (approved):
LGTM

PR #1266, opened 2023-08-18
💄 UX - Removed Duplicate Action When Setting Employment Status

### LINKED ISSUE #1157: 🐛 Employment Opportunity - Adding in Tina doesn't update UI
**Describe the bug**

**To Reproduce**
1. Go to https://www.ssw.com.au/admin/index.html#/~/employment
2. Add an opportunity
4. Tina editor shows the new role, but the UI does not

**Expected behavior**

image

```[tasklist]
### Tasks
- [x] Verify the useEffect change in the comment below fixed the display issue
- [x] Then refactor to remove the status value from the opportunities schema - have the status be determined by the ranking of opportunities on the employment page
- [x] Rewrite documentation w/ new functionality + redo done video
```

Harry-Ross:
* Removed "status" attribute from opportunities schema, replaced with selection on the /employment page (was already being done, simplified process)

image
**Figure: What now determines if a position is filled or not**

* Simplified filtering code on employment page
* Unified types and consts for job status
* Made animations quicker on employment page

Fixes #1157

Affected routes:

- `/employment`


bradystroud (changes requested):
Hey @Harry-Ross 
Really excited for this improvement 🙂 

I encountered a bug when I was trying to add an opportunity to the list
 
image

Since the process has changed, will we need to update docs + film a new video?

components/filter/opportunities.tsx
```diff
@@ -71,11 +74,11 @@ export const Opportunities = ({ opportunities }: OpportunitiesProps) => {
       ]}
     >
       {selectedStatus !== 1 &&
-        !!filteredOpportunities.find((o) => o.status === jobStatus[0]) && (
+        filteredOpportunities.some((o) => o.status === "Available") && (
```
bradystroud: can we avoid the hardcoded strings?

pages/employment/index.tsx
```diff
@@ -28,39 +32,24 @@ export default function EmploymentPage(
...
-  const removeExtension = (file: string) => {
-    return file.split(".")[0];
-  };
+  const opportunities: OpportunityType[] =
+    data.opportunitiesConnection?.edges.map((o) => {
+      const status = data.employment.opportunities.some(
+        (chosen) => chosen?.opportunityRef?.title === o?.node?.title
+      )
+        ? "Available"
+        : "Filled";
```
bradystroud: More hardcoded stuff 😕

bradystroud (changes requested):
@amankumarrr Identified a problem with this change.

What if I want to add a position, but for it not to be on the Employment page straight away?
It will appear in filled (when its not)
Another issue is when a position is filled in one location, but not another - right now you would mark one as filled, create a new one with the not filled location.

I think this is fine for now, as it is a good UX improvement. These concerns should be moved to a new issue.

bradystroud (approved):

PR #1304, opened 2023-08-24
Call Us - for desktop and mobile links 

### LINKED ISSUE #1300: ✨ SSW Website - Call us link
@2fernandez 
**Description:**
Currently, the 'Call Us' button on mobile directly initiates a phone call when clicked, while on desktop, it prompts users to open a compatible application for making the call. To align the behavior, we aim to have the button open the '[Contact Us](https://ssw.com.au/company/contact-us)' page on desktop while maintaining the existing behavior on mobile devices.

**Acceptance Criteria:**
Both scenarios should be a normal hyperlink 

CallUS
**Figure: Call us button on landing page**

amankumarrr:
Affected routes: 
-  `/*`

Fixed #1300


components/util/socialIcons.tsx
```diff
@@ -103,7 +103,18 @@ export interface SocialIconsParams {
   excludeMobile?: SocialTypes[];
 }
 
+export const isMobileDeviceAgent = () => {
+  return /Mobile|iP(hone|od|ad)|Android|BlackBerry|IEMobile/.test(
+    window.navigator.userAgent
+  );
+};
+
```
Harry-Ross: Can you move this to a file in /services - so its easier for devs to see they can use it

amankumarrr: I just replaced this function with a library named `react-device-detect` , as here we were using regex to test and it is not reliable as well as testable.

bradystroud (changes requested):
As discussed with @jaydenalchin this button should say 'contact us' when its on desktop

2fernandez:
As per my conversation with @amankumarrr.

- [x] Can all (internal links) Contact us / Call us link open in the same browser window
- [x] Can all social media links remain the same : they open in a new browser window

Thank you.

bradystroud (approved):

PR #1330, opened 2023-08-31
Updating CORS policy in NextJs API

### LINKED ISSUE #1331: CORS - Update Policy to get request from Rules and People website
We need to allow SSW.People and SSW.Rules to send their payload to SSW.Website, due to CORS policy, our server is not allowed to talk to a server on a different port.

amankumarrr:
fixes: #1331


Harry-Ross (approved):
lgtm 🚢

amankumarrr:
This issue can be closed since it's no longer required. We considered this as a contingency to implement the necessary CORS policy, enabling communication with both SSW.People and SSW.Rules. However, since all three sites share the same origin, there is no need to add a CORS policy.

PR #1355, opened 2023-09-06
🐛 Bug - Home Page - Rules Link prefetch returns 404

### LINKED ISSUE #1344: 🐛Bug - Home Page - Rules Link prefetch returns 404
### Background

This was identified, while investigating the sources of big number of failed requests in Application Insights.

The error is thrown by the prefetch request of https://www.ssw.com.au/rules/.

Next.js Link component performs a prefetch request on hover of any links that are relative or absolute with the same domain.

Because [Rules website](https://www.ssw.com.au/rules/) and [SSW website](https://www.ssw.com.au/) have the same domain, Next.js Link tries to prefetch Rules.

However, because Rules is actually its own repository, Next.js does not generate a next/data/..../rules.json file. 

Because that file does not exist, but the prefetch still happens, 404 error is logged.

### Steps to replicate:

1. Go to https://www.ssw.com.au/
2. Hover over the Rules link card
3. Open Dev Tools
4. Notice Error logged in the console
5. Open network tab
6. Notice request to rules.json fail

Image
Figure: Seems like the Link for Rules is the source of error.

### Solution

Let's convince Next.js Link that it does not need to prefetch Rules link

vladislav-kir:
As per the issue #1344, prefetch requests to SSW Rules link by the Next.JS link component always return 404, because the next/data files are not generated for /rules page, since it comes from a different repo.

Changing the link in the card for Rules page from:
    https://www.ssw.com.au/rules
To:
    https://ssw.com.au/rules
    
Since Next.JS link will only prefetch the pages with the same domain, I am hoping this is enough to stop prefetching, and so the constant 404 errors

- Affected routes: https://www.ssw.com.au/

- Fixed #1344 

image
Figure: Link for Rules is the source of error.

### DIFF
--- a/content/pages/home.mdx
+++ b/content/pages/home.mdx
@@ -73,7 +73,7 @@ beforeBody:
         color: darkgray
         imgSrc: /images/service-cards/Expert.png
       - title: Rules
-        link: 'https://www.ssw.com.au/rules'
+        link: 'https://ssw.com.au/rules/'
         color: darkgray
         imgSrc: /images/service-cards/Rules.png
     links:
--- a/content/pages/home.mdx
+++ b/content/pages/home.mdx
@@ -73,7 +73,7 @@ beforeBody:
         color: darkgray
         imgSrc: /images/service-cards/Expert.png
       - title: Rules
-        link: 'https://ssw.com.au/rules/'
+        link: 'https://www.ssw.com.au/rules'
         color: darkgray
         imgSrc: /images/service-cards/Rules.png
     links:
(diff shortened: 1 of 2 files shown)


content/pages/home.mdx
```diff
@@ -73,7 +73,7 @@ beforeBody:
         color: darkgray
         imgSrc: /images/service-cards/Expert.png
       - title: Rules
-        link: 'https://www.ssw.com.au/rules'
+        link: 'https://ssw.com.au/rules/'
```
bradystroud: I think its better that we change the component to use <a> instead of <Link>
Theres no benefit of `<Link>` for external links

bradystroud (changes requested):

content/pages/home.mdx
vladislav-kir: I changed small cards to use a regular anchor tag, instead of <Link>, and reverted the Rules link.

Those 4 small cards will no longer prefetch. any other cards will not be affected

bradystroud (approved):

PR #1358, opened 2023-09-07
Create lead - adding doc reference

### LINKED ISSUE #1332: 📄Create Lead - Document it in SharePoint
To: @pierssinclairssw 
CC: @Marxoz  @wicksipedia @adamcogan  @Aibono1225  @2fernandez @camillars @samsmithssw @bradystroud @JackDevAU 
Subject: RE: ⚠️[URGENT] - Leads through SSW.Rules and SSW.People

We are currently using the same endpoint (i.e., api/create-lead) in three different places (SSW.Webiste, SSW.Rules and SSW.People) for lead creation. This situation poses a risk, as changes made by a new developer in one location could potentially affect the other areas. To mitigate this, it's essential to thoroughly document the usage to ensure that any modifications are made carefully and do not cause unintended issues.
```[tasklist]
### Tasks
- [x] Document it in SharePoint
- [x] Add the link to that in the code where we are calling out that API in all three places
```

amankumarrr:
- Affected routes: 
 `/api/create-lead`

- Fixed #1332 

Document is test passed by @christianmorfordwaitessw

### DIFF
--- a/pages/api/create-lead.ts
+++ b/pages/api/create-lead.ts
@@ -19,6 +19,8 @@ export default async function handler(
   try {
     if (req.method === "POST") {
       const { Recaptcha } = req.body;
+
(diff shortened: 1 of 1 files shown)


bradystroud (approved):
Not a fan of putting a private link (only SSW can read it) on public code...

I cant think of a better way to do this so approving

PR #1394, opened 2023-09-14
Fixing - Toggle functionality on map for offices

### LINKED ISSUE #760: 🐛  Home Page | Contact Us - NSW shows Sydney but not Newcastle
**Describe the bug**

image

**To Reproduce**
1. Go to home page
2. Scroll to `Contact Us`
3. Click NSW state
4. Only Sydney is expanded, not Newcastle 😕

**Expected behavior**
Syd and Newcastle states are expanded

amankumarrr:
- Affected routes: `/` 

- Fixed #760 

ezgif com-optimize

**GIF: Toggles on hovering map and tabs**

### DIFF
--- a/components/blocks/aboutUs.tsx
+++ b/components/blocks/aboutUs.tsx
@@ -97,6 +97,9 @@ export const AboutUs = ({ data }) => {
     defaultOffice?.addressRegion || ""
   );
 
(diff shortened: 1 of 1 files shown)


bradystroud (approved):
For now this is a good enough solution, but I think we should redesign this in the future

PR #1489, opened 2023-10-04
Update carousel ug image

Geordie88:
- Affected routes: 

Updating the carousel Banner to be the up to date image (current one is for September user group


Harry-Ross:
Hi Geordie, is this using the same find out more button as was on the other carousel images?

Also, can you change the typeface of the title and .NET UG logo to be bold, with same font settings as Luke's image from last month? It would be better if it was more consistent with the other images in the carousel. 

image

Harry-Ross (approved):
lgtm 🚤

PR #1521, opened 2023-10-10
📄 Marketing assistant position - mark as filled

tiagov8:
As per @PennyWalker request


bradystroud (approved):
LGTM
Thanks for also adding a title 🙂

PR #1532, opened 2023-10-12
New mega menu

### LINKED ISSUE #65: ♻️ Megamenu | Replace with a good menu
_**Pain:**_ The NPM package poses a ton of issues for the NextJs site. The way to fix was to import the Megamenu into the lib folder and update it to work with TypeScript (Fix the image displaying).

This is a bunch of hoops that we need to jump through in order to drag this terrible menu around into every single new project we make.

**_Proposed Solution:_** 

### **Option 1 (Strongly recommended):** Re-design and re-implement the SSW.Megamenu.
✅✅✅ Re-designed - Feels fresh and has better UX
✅ Performance - Better performance for all Websites and platforms
❌ Cost - It will take time to approve a new menu design and implement.

### **Option 2 (Discouraged):** Update the existing SSW.Megamenu to be properly compatible with all projects we want it on.
❌❌❌ Cost - Far too time consuming to refactor
❌❌ Poor UX - The Megamenu has been allowed to be far too weighty for far too long. We can do better.

### **Option 3 (Do nothing, (Workaround)):** Continue to use the Megamenu as is.
❌❌❌ Poor UX - As Above
❌❌❌ Poor DevX - Implementing the Megamenu is a nightmare for developers (I've done it twice). It is never as easy as it should be. It is terrible. And feels terrible.
❌❌❌ Cost - You'd think that doing nothing means we don't waste time implementing it. But that isn't true. There's a plethora of issues and inconsistencies in the Megamenu that make it nearly impossible to easily implement.
❌❌❌ The menu causes problems on all internal projects, not just this one

_Originally posted by @tkapa in https://github.com/SSWConsulting/SSW.Website-v3/issues/63#issuecomment-1403011764_

wicksipedia:
Implementing a v1 of the new mega menu

- Affected routes: all routes as this is in the layout root

- Fixes #65 

CleanShot 2023-10-12 at 17 55 25

CleanShot 2023-10-12 at 17 55 34

CleanShot 2023-10-12 at 17 55 43


bradystroud (approved):
LGTWilliam

PR #1562, opened 2023-10-19
re-use docker to build image so that sharp can resize images

### LINKED ISSUE #1461: 💸 Tech Debt - webP not being used for image optimization for prod
It was observed that webP is being served for images during local dev but not prod

This affects pagespeed insights

```[tasklist]
### Tasks
- [ ] Ensure Image Optimization API is being used for images e.g. Tina, old SSW website (biggest improvements here)
```

wicksipedia:
See https://nextjs.org/docs/messages/sharp-missing-in-production

Docker container was not using `sharp` as required by nextjs.
Modified build process and dockerfile to use `sharp`
Should have a minor improvement on performance (as webp files are smaller)

- Fixes #1461 

- Affected routes: *

CleanShot 2023-10-20 at 11 47 05@2x


bradystroud (approved):
LGTM

PR #1582, opened 2023-10-20
disable static builds on prs

wicksipedia:
PR builds take about 15mins.... this is too long

Added a flag to disable static generation on PRs to expedite reviews.

- Affected routes: N/A


Harry-Ross:
Because there's no static generation, does that mean that we don't have the catch-all build for every page? Doesn't that mean that potential bugs could go unnoticed with large changes (e.g. schema) unless we actually went to that page?

Harry-Ross:
Adding to my previous point - I've found the build process going over every page's getStaticProps often catches something that I didn't on my dev server - feels like the deployments are now very similar to a dev server by generating dynamically, and the errors would come up in production instead.

wicksipedia:
This is a draft PR for experimentation purposes only

wicksipedia:
this change halved the build time but we need ISR for real improvmeents

PR #1634, opened 2023-10-31
Fix - Broken image for upcoming events

### LINKED ISSUE #1633: 🐛 Broken image - Upcoming Events
**Describe the bug**
In the case of upcoming events, if there is no associated image, a broken image placeholder is displayed.

**Expected behavior**
It should handle the case when there is not image associated. 

**Screenshots/Video**
image
**Figure: Broken image for .NET CONF**

amankumarrr:
- Affected routes: `/`

- Fixed #1633  

This fix handles if the image Path contains multiple dots or if there is not associated image with the events. 

⚠️ It doesn't handle if the image is giving 404 error

image

**Figure: Hiding img if it has not a valid path**


bradystroud (approved):
Looks good

PR #1665, opened 2023-11-06
Fix - Flickering data for Event's dates

### LINKED ISSUE #1664: 🐞 BUG - Events - shows incorrect start/end dates (especially on facebook ads)
When I look at the event page via a facebook ad - it shows the wrong dates for Melb & Syd (should be 13-14 & 16-17)

Image

When I go to the page directly, it shows the right dates, but if I refresh.... I can see the dates flicker from wrong to right

Image
**Figure: flicker**

amankumarrr:
Affected routes: 
 `/events/*`

- Fixed #1664

TODO


bradystroud:
Closing as this does not fix the issue

PR #1707, opened 2023-11-14
UX Designer position filled

tiagov8:
Marked position as filled as per @PennyWalker 's **Re: User Experience Expert & Web Designer - Remember to mark as "filled"**


bradystroud (approved):
LGTM

PR #1746, opened 2023-11-22
🚑 SEO - Fix Noindex Cookie

### LINKED ISSUE #1745: 🚑 Middleware - Breaking SEO
image

Currently, we are erraneously sending a noindex cookie

```[tasklist]
### Tasks
- [x] Fix to have noindex cookie only on staging and PR
- [x] Add tests to ensure cookies are applied correctly in PR, staging environments
```

Harry-Ross:
- Affected routes: *

- Fixed #1745


ui-tests/seo-main.spec.ts
```diff
@@ -0,0 +1,9 @@
+import test, { expect } from "@playwright/test";
+
+test("No index header not present", async ({ page }) => {
+  const response = await page.goto("/", {
+    waitUntil: "networkidle",
+  });
+  expect(response.ok()).toBeTruthy();
+  expect(response.headerValue("X-Robots-Tag")).not.toEqual("noindex");
```
wicksipedia: should this instead check that there is no `X-Robots-Tag` set?
it may make the intent more clear

Harry-Ross: Done - as per https://playwright.dev/docs/api/class-request#request-header-value I am now checking if the header value is falsey (null)

ui-tests/.gitignore
```diff
@@ -1,7 +1,6 @@
 node_modules/
 /test-results/
 /playwright/.cache/
-/playwright-report
```
wicksipedia: wont't this mean that people will commit the reports when they run tests locally?

there was a massive index file checked in as part of this pr

CleanShot 2023-11-24 at 11 50 20@2x

Harry-Ross: We need to keep this not gitignored so when the Playwright test is run in a GH Action, we can see the result. 

However, you are right that the report I ran shouldn't be added - I removed the HTML file from the PR.

ui-tests/seo-main.spec.ts
```diff
@@ -0,0 +1,12 @@
+import test, { expect } from "@playwright/test";
+
+test("No index header not present", async ({ page }) => {
+  await page.on("response", async (response) => {
+    expect(await response.headerValue("X-Robots-Tag")).toBeFalsy();
+  });
```
wicksipedia: this seems so backwards, can you log an item to try and improve this?

wicksipedia (approved):
lgtm

PR #1773, opened 2023-11-29
uploading thumbnail for DDD conference event

sethdaily:
As per email: RE: DDD Brisbane 2023 - Accepted presenters and their talks


ae-cher-ssw (approved):
Looks good to me!

PR #1799, opened 2023-11-29
Remove some text-xxs usages

### LINKED ISSUE #1147: 💸 Tech Debt - Increase base font-size
cc @BrookJeynes @bradystroud @amankumarrr 

**Description**  

Usually it's [recommended to use a font-size at least 16px](https://www.learnui.design/blog/mobile-desktop-website-font-size-guidelines.html) which is exactly what we have at ssw.com.au

However I've received some feedback from @adamcogan considering it too small (Email subject: **Home page banner**)

> Can you increase the font size?

Let's make it 18px instead.

I also think we don't need to set a different base line-height, keeping the default

**Proposed solution**  
1. Please update the CSS replacing

```
html {
    line-height: 1.5;
    ...
}
```

with

```
html {
    font-size: 18px;
    ...
}
```

**Note:** This change will affect the website content pretty much everywhere. We may need to adjust some font sizes here and there (e.g https://github.com/SSWConsulting/SSW.Website/issues/1148) since they are all based in this piece of css

---

image0

**Figure: Font size on the bottom red box seems too small to Adam**

wicksipedia:
.nvmrc - aligns version with `Dockerfile`
HTML - Removes usages of `text-xxs` except in footer

- Affected routes: `/` and `/events`

- Fixed #1147

CleanShot 2023-11-30 at 09 44 25

CleanShot 2023-11-30 at 09 44 50


jeoffreyfischer (approved):
LGTM

PR #1971, opened 2023-12-28
♻️ Megamenu - Refactor SocialIcons

### LINKED ISSUE #1560: 💸 Megamenu - export it as a component
**Description**  
In order to get the new megamenu out sooner, we shipped a version that is only built on the main website.

This needs to be able to be used on `/people` and `/rules`

**Proposed solution**  

1. This is blocked by #1559
2. Add an api endpoint to retrieve the layout as configured in Tina
3. Move the megamenu code to `/lib`
4. Megamenu - if layout data prop is not passed in, go to api endpoint to retrieve layout data (maybe store in local storage for 1 hour so that we don't have to call it all the time)
5. Megamenu - add a new prop to generate the link (remove the dependency on next/link since rules and people use gatsby - they should inject their own implementation)
6. Publish megamenu as a new npm package that can be picked up by people and rules

```[tasklist]
### Tasks
- [ ] https://github.com/SSWConsulting/SSW.Rules/issues/1188
- [ ] https://github.com/SSWConsulting/SSW.Website/issues/2115
```

Harry-Ross:
* Cleaning up `<SocialIcons>`

- Affected routes: `/`

- Fixed #1560


amankumarrr (approved):
LGTM

PR #2028, opened 2024-01-11
Adding image for the Newsletter SSWTV

PennyWalker:
Adding image for the Newsletter SSWTV as per standard required for me


Geordie88 (approved):
I guess I'll allow it :D

PR #2058, opened 2024-01-15
Updating JK torso image

sethdaily:
Replacing with a better one.
As per email: Brisbane UG - Updates necessary for new speaker


bettybondoc (approved):
LGTM 🚀

PR #2149, opened 2024-02-14
💄Events - Increasing events counts and fixing location

### LINKED ISSUE #2128: 🐛 SSW Events page adds SSW Chapel as a link as default for locations
URL: https://www.ssw.com.au/events?tech=all&type=all

Subject: 🪲 SSW Events page adds SSW Chapel as a link as default for locations ❌
FROM: @camillars 
CC: @wicksipedia @sethdaily 

**Describe the bug**
>Every now and then we sponsor events and add them this page. However, these are not hosted/run by SSW e.g. NDC or DDD, and if we select  ‘Sydney’ as a location, by default it will add [SSW Chapel Sydney](https://sswchapel.com.au/Sydney/) as a location.

> 1.	Can we fix this so it doesn’t behave this way for external events? Perhaps we could add ‘Sydney’ and ‘[SSW Chapel Sydney](https://sswchapel.com.au/Sydney/)’ as different options so it’s not confusing.

**Screenshots/Video**
SSWChappenlSydney

**Figure: the locations currently point to an SSW Chapel but if it’s an external event, they shouldn’t**

### LINKED ISSUE #2145: 💄 UX - Not Enough Events Showing
cc @camillars 

As per Camilla's email, the 15 events shown on https://www.ssw.com.au/events isn't enough. 

```[tasklist]
### Tasks
- [x] Increase the number of events displayed to 20
```

amankumarrr:
Fixed #2145 & fixes #2128 

Affected Route: `/events`

Also fixed the some UI after getting feedback from @jaydenalchin 

Sreenshot:
image

**Figure: Events page after minor tweaks**


jaydenalchin (approved):

camillars:
Hi Aman,

as per our conversation just now, test passed - this is exactly what I wanted. Thank you 🌷🎉

tailwind.config.js
```diff
@@ -226,6 +226,8 @@ module.exports = {
         sswRed: "#cc4141",
         sswDarkRed: "#8e2c2c",
         sswBlack: "#333333",
+        sswGray: "#797979",
```
Harry-Ross:
```suggestion
```

We already have a `ssw-gray` variable identical to this below, better to use that instead

tailwind.config.js
```diff
@@ -226,6 +226,8 @@ module.exports = {
...
+        sswDarkGray: "#666666",
```
Harry-Ross: Is this an official SSW color? (i.e. from https://www.ssw.com.au/ssw/logo/ssw/) If not just get rid of it and use `gray-650` (duplicate value)

tailwind.config.js
amankumarrr: right, I just removed it. 

thanks

tailwind.config.js
amankumarrr: I've got it from the TimePro, and I asked @jaydenalchin about it. I think we should use something like ssw-gray-dark for this instead, what do you think?

Harry-Ross (approved):
lgtm 🚢

PR #2166, opened 2024-02-19
App Insight - Adding dependency tracking 

### LINKED ISSUE #2152: 🐛 App Insight - Not logging client side error
**Describe the bug**
It seems like our client side errors not logged with App Insight anymore.

**Expected behavior**
It should log both server and client side errors.

**Screenshots/Video**
image

**Figure: App Insight's log for the browser**

amankumarrr:
- Fixed #2152


wicksipedia (approved):
From the azure portal, I can see failures being logged now - so it is an improvement.
We may need to look at the performance blade as it is only reporting dependency data

PR #2324, opened 2024-03-19
Logo - Adding FireBootCamp page

### LINKED ISSUE #2278: 📄Migration | Logo | FBC page
URL: https://ssw.com.au/ssw/logo/FireBootCamp/

- [x] Migrate the page  
- [x] Add FrontDoor redirects

amankumarrr:
Affected Route: `/logo/firebootcamp`
Fixed: #2278

**Screenshot:** 
image

**Figure: Logo - Firebootcamp page**


components/util/constants.tsx
```diff
@@ -169,3 +169,9 @@ export const sswColors = {
   "#AAAAAA": "bg-ssw-gray-light",
   "#797979": "bg-ssw-gray",
 };
+
+export const CustomClasses = {
```
Harry-Ross:
```suggestion
export const customClasses = {
```

Maybe also make it specific to margin i.e. `customMarginClasses`

components/blocks/fixedColumns.tsx
```diff
@@ -7,34 +7,43 @@ import { componentRenderer } from "./mdxComponentRenderer";
...
 
 export const fixedColumnsSchema: Template = {
   name: "FixedColumns",
   label: "Fixed Column Layout (2 columns)",
   fields: [
+    {
+      type: "string",
+      label: "Header",
+      name: "header",
+    },
```
Harry-Ross: Should this be a rich-text field? Or even just moving that specific header to another block instead of this one. It seems odd to include this as a prop.

components/util/constants.tsx
amankumarrr: I'm thinking of having some more classes here in the future to have more customization, currently, it only contains margin but we might be adding some padding as well

components/blocks/fixedColumns.tsx
amankumarrr: I updated the type to rich-text and renamed it to 'headerSection' to add more meaning. It's beneficial to have it this way because it ensures consistent padding within its fixed layout block.

Harry-Ross (approved):
LGTM 🚣‍♂️

PR #2456, opened 2024-04-11
🎙️ Updating Angular 2day workshop in Melbourne date

camillars:
As per my chat with @PennyWalker, we had to change the date of the Melbourne event so it doesn't clash with another internal event.

- Affected routes: https://www.ssw.com.au/events/angular-workshop


AshAnil (approved):
LGTM

PR #2531, opened 2024-04-22
Increase image quality in carousel on the home page

johnxu16:
- Affected routes: 
- /

- Fixed #2502 

image

- [ ] If adding a new page, I have followed the [📃 New Webpage](https://github.com/SSWConsulting/SSW.Website/issues/new?assignees=&labels=&projects=&template=new_webpage.yml&title=%F0%9F%93%84+%7B%7B+TITLE+%7D%7D+) issue template

- [x] Include done video or screenshots

PR #2593, opened 2024-05-06
Home - Fixing Banner height

### LINKED ISSUE #2588: 🐛 Home - 💄Banner is cutting off
**Describe the bug**
Following the recent update, the height of the banner has been altered, causing the image to be cut off from the top. As per my conversation with @wicksipedia and @bettybondoc, we've decided that we need to either correct the height or revert the change responsible for this issue.

- [x] Fix the issue 

**Screenshots/Video**

Image

**Figure: Penny is being cut off in the banner**

amankumarrr:
Fixed: #2588

Affected Route: `/`

Reverted the changes made by this pr - https://github.com/SSWConsulting/SSW.Website/pull/2531, we can't use `fill` with `object-fit:cover` as it will crop the image and as a result, some objects in the image will be cut off. 

✅ Also added Turbo in config to improve the local development
✅ Reducing the icons size to reduce the image load 
 
**Screenshot:**

image

**Figure: Banner image with fixed height and width**

### DIFF
--- a/components/blocks/carousel.tsx
+++ b/components/blocks/carousel.tsx
@@ -98,6 +98,7 @@ const CarouselItemImage = (props: CarouselItemImageProps) => {
         alt={label}
         height={388}
         width={1080}
+        sizes="100vw"
         priority={index === 0}
       />
       {/* `legend` required so that the carousel works properly */}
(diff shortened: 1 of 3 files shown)


andrewwaltosssw (approved):
lgtm 🚢

PR #2761, opened 2024-06-20
2653 add event eoi

### LINKED ISSUE #2653: ✨ Events - add EOI form to capture user interest
cc @uly1 @Levijj22 @camillars @sethdaily @landonmaxwell 

For events that we run (but there is no date for the next event) - we need to be able to grab expressions of interest.

**Describe the solution you'd like**

Ideally, on an event page, instead of showing the cities/dates - there would be a form that users can submit and we can get back to them when there is a firm date.

```[tasklist]
### Tasks
- [x] Jotform - Create JotForm to capture details (including url that the form is on - this form should be generic and not just for events)
- [x] CRM - Check with sales/marketing if we want the user details added to CRM/Newsletter
- [x] TinaCMS - Add setting for events to either set dates or have EOI form
```

Image

Calinator444:
cc: @sethdaily, @camillars 

### Description
This PR adds a re-usable registration of interest form as a Tina component for the SSW Website. The form captures the user's details as well as a link to the event they were interested in. As per my conversation with @sethdaily the details are currently being stored in the JotForms submission data. In addition, a power-automate flow also registers users for our newsletter.

Fixed #2653 

### Note
The power automate flow for adding these users to our contacts hasn't been set up yet. This form will still capture registrations of interest in JotForm but until my flow hit's prod none of the submissions will be added to crm. 

### Screenshots

ROI button Tina
**Figure**: **Registration of Interest button being edited in Tina**

Registration of Interest Form
**Figure**: **Registration of interest form**

image
**Figure**: **Registration of interest form in Tina**


components/events/interestForm.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <JotFormEmbed
+      jotFormId={`${eoiFormId}?pageUrl=${props.pageUrl}`}
+      buttonText={props.buttonText || "I am interested"}
+      containerClass=""
+    />
+  );
+};
+
+const eoiFormId: string = "241638083078865";
```
bradystroud: It would be better to not hardcode this

jeoffreyfischer:
The following check has failed - please investigate:
PR - Lint PR/ Check linked issues

Calinator444:
> The following check has failed - please investigate: PR - Lint PR/ Check linked issues

**Done**: I made sure the issue was linked correctly then re-ran the job

bradystroud (changes requested):

components/events/interestForm.tsx
wicksipedia: agree, should be in global config

Calinator444: Done - I've moved it to the global config

bradystroud (approved):

PR #2822, opened 2024-07-10
Fix event images

### LINKED ISSUE #2694: 🚶 Events - migrate SharePoint list data to Tina
Currently the Events are sourced from a SharePoint list - this isn't ideal since:

- we need to query an external source
- it makes it harder to create a new event
  first add an image to the website, then add data to a list
  if you do it too quickly, it can cause tests to fail on deployment due to broken images
- this will help perf stats as next will be able to optimise them

```[tasklist]
### Tasks
- [x] Create `event` collection
- [x] Write script to export SharePoint event list and presenter data
- [x] Standardise event thumbnail location (in schema)
- [x] Reference TinaCMS event data for events and Live UG
- [x] Remove unused environment variables from build
- [x] Update marketing standard (give Camilla a video)
```

Calinator444:
### Description 

A number of the thumbnail images on events are broken because the events use urls that are not in Tina. In this PR I've updated the the events to point to files stored in this repo. 

Fixed #2694

- added fallback for events to use the file from the website rather than tina cloud on 404
- Updated events to use files in the repo


content/events-calendar/hotfix.js
wicksipedia, line 1: given that the urls have been migrated - do we need this file?

components/filter/events.tsx
```diff
@@ -214,6 +214,15 @@ interface EventProps {
...
 
 const Event = ({ visible, event, jsonLd }: EventProps) => {
+  const [thumbnail, setFallbackImage] = useState(event.thumbnail);
+  const handleImageError = () => {
+    const tinaUrl = /https:\/\/assets\.tina\.io\/[^/]+\/(.*)/;
+    const match = event.thumbnail.match(tinaUrl);
+    if (match) {
+      setFallbackImage(`/images/${match[1]}`);
+    }
+  };
```
wicksipedia, line 230: given that the thumbnail urls have been migrated... do we need this?

I'd fallback to a static image instead of rewriting the url

Calinator444: The point to files in the repo but not all of the files have been synced to Tina cloud. I've had issues syncing the repo with Tina cloud because of the size of the repo.

content/events-calendar/hotfix.js
Calinator444: My mistake, didn't mean to commit that.

Done - I've removed it

wicksipedia (approved):

PR #2828, opened 2024-07-15
✨ Events - new API endpoints (/past + /upcoming)

### LINKED ISSUE #2785: Events - add endpoint to return upcoming events
Discovered whilst migrating SSW.People to blob storage - this will block switching off the VM

There is a call to the v1 website that returns upcoming events - https://github.com/SSWConsulting/SSW.People/blob/f81d9de4965d0fd3622b4befdad302aa2e1a2237/src/helpers/eventHelper.js#L4

We need a similar endpoint on v3

### Tasks
- [x] Create an API endpoint that can query the most recent events
- [x] add the route `api/events/upcoming/:presenter` for getting upcoming events by presenter
- [x] add the route `api/events/past/:presenter` for getting past events by presenter

### Acceptance criteria
- [x] Endpoints are implemented using the app router

Calinator444:
@wicksipedia 

### Description 
This PR adds endpoints designed to replace the graph API used to fetch events from sharepoint

### Affected routes: 

- /api/events/past
- /api/events/upcoming

**Note**: both of these endpoints have a required `presenterName` query parameter

- Fixed #2785

### Important note
because the naming scheme for this endpoint is and parameters required is different than the existing endpoint used for SSW People, their team will need to switch to our new APIs after this PR is merged.

The upcoming and past events list on people is also broken right now. To produce the screenshot I fixed the events list on my local instance of SSW People without renaming any of the expected parameters.

Also, this API only includes the data necessary for SSW People to render correctly. I'm unsure if the endpoint is used in existing SSW Products If so there may be breaking changes because I haven't migrated every field from the old endpoint, only those necessary for SSW People to render correctly.

### Screenshot

image

**Figure**: **Goldie's people page being populated with data from the new events APIs**

image

**Figure**: **Wicksy's people page being populated with data from the new events APIs**


app/api/events/past/route.ts
```diff
@@ -0,0 +1,20 @@
...
+  let presenterName = req.nextUrl.searchParams.get("presenterName");
+  if (!presenterName)
+    return new Response("presenterName is required", { status: 400 });
+  presenterName = hyphenateUrl(presenterName);
+  const eventClient = await client.queries.getPastEventsQuery({
+    fromDate: new Date().toISOString(),
+    top: 999 /* GraphQL will limit the count to 50 by default, 
+    but were doing the filtering outside of the query because GraphQL has no "like" filter
+    https://tina.io/docs/graphql/queries/advanced/filter-documents/#operator-types
+    */,
```
wicksipedia: add a TODO and link to the issue to clean up the tech debt

wicksipedia:
when I query `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/upcoming` or `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/past`- it says `presenterName is required`

I would've expected it to return the list of events, is that expected?

app/api/events/getEvents.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const fixRelativeUrl = (url: string) => {
+  if (url.startsWith("/")) {
+    return `https://ssw.com.au${url}`;
+  }
+  return url;
+};
+
+const descriptionToPlainText = (description) => {
```
wicksipedia: it looks like this is formatting html and not plain text

Calinator444:
> when I query `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/upcoming` or `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/past`- it says `presenterName is required`
> 
> I would've expected it to return the list of events, is that expected?

Yes, I added a 400 response when presenterName isn't provided.

app/api/events/getEvents.ts
```diff
@@ -0,0 +1,69 @@
...
+  };
+};
+
+export const getEventsWithClient = async (eventClient, presenterName) => {
+  const events = [];
+  // TODO: remove client side after fixing events with multiple presenters in the name
+  for (const event of eventClient.data.eventsCalendarConnection.edges) {
+    if (
+      event.node.presenterName &&
+      event.node.presenterName.includes(presenterName)
```
wicksipedia: it looks like this check is case sensitive i.e. hitting `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/upcoming/?presenterName=matt-wicks` returns `[]`

app/api/events/past/route.ts
```diff
@@ -0,0 +1,20 @@
+import { hyphenateUrl } from "app/api/hyphenateUrl";
+import { NextRequest } from "next/server";
+import client from "../../../../tina/__generated__/client";
+import { getEventsWithClient } from "../getEvents";
+
+export async function GET(req: NextRequest) {
+  let presenterName = req.nextUrl.searchParams.get("presenterName");
+  if (!presenterName)
+    return new Response("presenterName is required", { status: 400 });
```
wicksipedia:
```suggestion
  if (!presenterName) {
    return new Response("presenterName is required", { status: 400 });
  }
  
```
braces for readability

app/api/events/upcoming/route.ts
```diff
@@ -0,0 +1,16 @@
+import { hyphenateUrl } from "app/api/hyphenateUrl";
+import { NextRequest } from "next/server";
+import client from "../../../../tina/__generated__/client";
+import { getEventsWithClient } from "../getEvents";
+
+export async function GET(req: NextRequest) {
+  let presenterName = req.nextUrl.searchParams.get("presenterName");
+  if (!presenterName)
+    return new Response("presenterName is required", { status: 400 });
```
wicksipedia:
```suggestion
  if (!presenterName) {
    return new Response("presenterName is required", { status: 400 });
  }
  
```
braces for readability

app/api/events/getEvents.ts
Calinator444: Done - I've made it case insensitive

app/api/events/getEvents.ts
Calinator444: I was trying to emulate the format the original API was using for this field, which would use divs an linebreaks to separate paragraphs. I figured I could achieve the same effect using `<p>` tags to create line breaks (by default Tina stores line breaks as separate "paragraph" props in the json file, I'm just converting it back to html).
image

wicksipedia: I was commenting on the name of the function since it doesn't do what it says on the box

Ideally, I'd avoid adding extra logic and just fix the data at the source - this can be done as a separate pbi

app/api/events/past/route.ts
Calinator444: Done - I added a todo and a PBI for fixing this

Calinator444:
> when I query `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/upcoming` or `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/past`- it says `presenterName is required`
> 
> I would've expected it to return the list of events, is that expected?

Calinator444:
> when I query `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/upcoming` or `https://app-sswwebsite-9eb3-pr-2828.azurewebsites.net/api/events/past`- it says `presenterName is required`
> 
> I would've expected it to return the list of events, is that expected?

Done - I've updated it to return the first 10 events if no `presenterName` is provided

app/api/events/past/route.ts
Calinator444: Done - I've updated the if statements in this file to use braces (I removed the 400 so this suggestion is no longer relevant, see reply https://github.com/SSWConsulting/SSW.Website/pull/2828#issuecomment-2232047081)

app/api/events/upcoming/route.ts
Calinator444: I've removed the 400 since this is no longer relevant (See comment https://github.com/SSWConsulting/SSW.Website/pull/2828#issuecomment-2232047081)


However I did update he if statements in this function to use braces.

app/api/events/getEvents.ts
Calinator444: Ok, I'll rename the function but I think fixing the function is a low priority since I can't see the prop being used anywhere on rules. We might even just be able to omit it from the API.

app/api/hyphenateUrl.ts
```diff
@@ -0,0 +1,3 @@
+export const hyphenateUrl = (name: string) => {
+  return name.replace("-", " ");
```
jeoffreyfischer: looks like your method does dehyphenateUrl instead of hyphenateUrl

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
+import client from "@/tina/client";
+
+export const getPastEvents = async (top, presenterName) => {
+  console.log("presenterName", presenterName);
```
jeoffreyfischer: no console.log in PRs

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+  );
+  return await fetchEventsWithClient(eventClient, presenterName, top);
+};
+
+const formatEvent = (event) => {
+  const url = event.url && fixRelativeUrl(event.url);
+  return {
+    Presenter: event.presenterName,
+    Url: {
+      Description: event.title, // this is alt text for the link, ask matt what he wants to do for it
```
jeoffreyfischer: I don't understand your comment "as matt what he wants to do for it"

jeoffreyfischer: Also comments should only be left in extreme necessity - this one seems to be sthg you should confirm with Matt. 
Once a decision is made, update the code and leave a comment in the PBI (and remove the comment from here).

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+export const fetchEventsWithClient = async (
+  eventClient,
+  presenterName: string | undefined,
+  top
+) => {
+  if (top) {
+    top = parseInt(top);
+  }
+  const events = [];
+  // TODO: remove client side after fixing events with multiple presenters in the name
```
jeoffreyfischer: If this TODO can be fixed in this PR, then do it and remove this comment.
If it can't be fixed here, then create a PBI and add the link to it in this TODO comment

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+      !presenterName ||
+      (event.node.presenterName &&
+        event.node.presenterName
+          .toLowerCase()
+          .includes(presenterName.toLowerCase()))
+    ) {
+      events.push(formatEvent(event.node));
+    }
+
+    if (events.length === (top || 10)) {
```
jeoffreyfischer: Create a const and assign it the value 10

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+      break;
+    }
+  }
+  return events;
+};
+
+const fixRelativeUrl = (url: string) => {
+  if (url.startsWith("/")) {
+    return `https://ssw.com.au${url}`;
```
jeoffreyfischer: Create a const for https://ssw.com.au

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+    });
+  });
+  return plainText;
+};
+
+const formatEventParams = (
+  top: string | undefined,
+  presenterName: string | undefined
+) => {
+  let topArg = 999;
```
jeoffreyfischer: why 999? is this really needed?
If yes, at least create a const

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+  if (top && !presenterName) {
+    topArg = parseInt(top);
+  } else if (!top && !presenterName) {
```
jeoffreyfischer: both your "if" statements have the same condition !presenterName.
Consider refactoring this code to sthg like:
if (!presenterName) {
    topArg = top ? parseInt(top) : 10;
  }

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+  top
+) => {
+  if (top) {
+    top = parseInt(top);
+  }
+  const events = [];
+  // TODO: remove client side after fixing events with multiple presenters in the name
+  for (const event of eventClient.data.eventsCalendarConnection.edges) {
+    if (
+      !presenterName ||
```
jeoffreyfischer: the logic inside your "if" should be extracted in its own method - this if statement is too complex.

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const fixRelativeUrl = (url: string) => {
+  if (url.startsWith("/")) {
+    return `https://ssw.com.au${url}`;
+  }
+  return url;
+};
+
+const descriptionToHTML = (description) => {
```
jeoffreyfischer: couldn't this be fixed at the source?

jeoffreyfischer (changes requested):

services/server/getEvents.ts
Calinator444: Done - I broke it off into a helper method

services/server/getEvents.ts
Calinator444: Done  - I broke it off into a constant and updated the method

services/server/getEvents.ts
Calinator444: Done - Added a link to the PBI for fixing this

services/server/getEvents.ts
Calinator444: Done - removed the console log

app/api/hyphenateUrl.ts
Calinator444: Done - I renamed the method to `dehyphenateUrl `

services/server/getEvents.ts
Calinator444: > why 999? is this really needed?

[There's no `LIKE %<statement>%` equivalent in GraphQL](https://tina.io/docs/graphql/queries/advanced/filter-documents/#operator-types) so I'm filtering by name on the back end See PBI for fixing this here here: https://github.com/SSWConsulting/SSW.Website/issues/2833
> If yes, at least create a const

Done - Added a const for this value

services/server/getEvents.ts
Calinator444: I asked Jack P over teams and he it can't be fixed at the source unfortunately.

services/server/getEvents.ts
Calinator444: Hmm, that actually makes a lot more sense. A nested if would do the same thing but this is cleaner

Done - Refactored presenter list override expression using a ternary operator

services/server/getEvents.ts
Calinator444: This field didn't get brought over from SharePoint so I'm deciding how I can mock it

jeoffreyfischer: I think you fixed it from "as" to "ask", so it makes sense now

services/server/getEvents.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const formatEvent = (event) => {
+  const url = event.url && fixRelativeUrl(event.url);
+  return {
+    Presenter: event.presenterName,
+    Url: {
+      Description: event.title, // this is alt text for the link, ask matt what he wants to do for it
+      Url: url,
+    },
+    title: event.title,
```
jeoffreyfischer: This should be an upper case "Title" to be consistent with the rest of the props

services/server/getEvents.ts
Calinator444: I discussed this change with @bradystroud and he's happy with it. I've removed the comment now.

Calinator444: Done - I discussed this with @bradystroud as Matt is away and he's happy with the change. I've removed the comment.

services/server/getEvents.ts
Calinator444: Done - I updated this prop to "Title"

services/server/getEvents.ts
bradystroud: As per our convo, check if people actually uses this, if not just remove it.

app/api/dehyphenateUrl.ts
bradystroud, line 1: move this

bradystroud (changes requested):

services/server/getEvents.ts
Calinator444: Done - I couldn't see it being used in the code on SSW People so I removed it

**Note**: Even though it's not being used anywhere in the code on people the service still expects this prop. I'll update the service after this.

Calinator444: Done - I couldn't see it being used in the code on SSW People so I removed it

Note: Even though it's not being used anywhere in the code on people the service still expects this prop. I'll update the service after this.

app/api/dehyphenateUrl.ts
Calinator444: Done - moved this method to the services folder

services/server/getEvents.ts
Calinator444: Done - moved assigned this to a constant

Calinator444:
@bradystroud I had a look on SSW People and it looks like Jernej's profile is still fetching properly even though his profile has a nickname. 

I suspect People is using the path to the profile to query the API.


image

bradystroud (approved):

jeoffreyfischer (approved):

PR #2845, opened 2024-07-22
Fix logo names

hveraus:
- Affected routes: /logo

- [ ] change the logo header with proper mode names. e.g. Dark Mode Color instead of Color on dark background


KristenHu (approved):
LGTM🎉

PR #2848, opened 2024-07-23
🔒Events - added access control allow origin header for the ssw poeple staging environment

### LINKED ISSUE #2785: Events - add endpoint to return upcoming events

Calinator444:
### Description

We need to test the events APIs in the staging environment for SSW People but the requests are being blocked by CORS since the staging environment is a different domain to SSW People,

- Fixed [#2785](https://github.com/SSWConsulting/SSW.Website/issues/2785)

### DIFF
--- a/app/api/events/past/route.ts
+++ b/app/api/events/past/route.ts
@@ -13,6 +13,8 @@ export async function GET(req: NextRequest) {
     status: 200,
     headers: {
       "Content-Type": "application/json",
+      "Access-Control-Allow-Origin":
+        "https://sapeoplestagjthgmptzb46i.z8.web.core.windows.net",
     },
   });
 }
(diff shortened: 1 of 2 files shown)


bradystroud (approved):
Not a fan of hardcoding the url, I'll approve this to test it but lets move it somewhere better later

PR #2866, opened 2024-07-29
🔓 modified access control header from past request to allow requests form ssw people and staging

### LINKED ISSUE #2831: 🗺️ Cloudflare - stop routing SSW.People through Front Door
This should be done once SSW.People is migrated to blob storage

Calinator444:
### Description

This will allow the staging URLs for SSW People and rules to 

- Fixed #2831 
### Screenshot

image


jernejk (approved):
LGTM. If we need to CORS lockdown the API, we can add this as deployment variables.

app/api/events/upcoming/route.ts
```diff
@@ -14,8 +14,7 @@ export async function GET(req: NextRequest) {
     status: 200,
     headers: {
       "Content-Type": "application/json",
-      "Access-Control-Allow-Origin":
-        "https://sapeoplestagjthgmptzb46i.z8.web.core.windows.net",
+      "Access-Control-Allow-Origin": "*",
```
wicksipedia: Don’t use *
Add the actual urls

app/api/events/past/route.ts
```diff
@@ -13,8 +13,7 @@ export async function GET(req: NextRequest) {
     status: 200,
     headers: {
       "Content-Type": "application/json",
-      "Access-Control-Allow-Origin":
-        "https://sapeoplestagjthgmptzb46i.z8.web.core.windows.net",
+      "Access-Control-Allow-Origin": "*",
```
wicksipedia: Don’t use *
Add the actual urls

wicksipedia (changes requested):

app/api/events/upcoming/route.ts
Calinator444: Done - I've updated the endpoints to return the appropriate access-control-allow-origin header only when the domain is trusted

app/api/events/past/route.ts
Calinator444: See comment https://github.com/SSWConsulting/SSW.Website/pull/2866#discussion_r1695130398

services/server/cors.ts
```diff
@@ -0,0 +1,11 @@
+export const setAllowOriginIfTrusted = (responseHeaders, origin) => {
+  if (trustedOrigins.includes(origin)) {
+    responseHeaders.headers["Access-Control-Allow-Origin"] = origin;
+  }
+};
+
+const trustedOrigins = [
```
jernejk, line 7: Move the configuration to `.env` so it can be later configured by the pipeline and ssw.com.au.

jernejk: We can leave this as is since it's probably going to be used by the same origin for the foreseeable future, and we just need this to support staging/production Blob Storage URLs.

wicksipedia (approved):

PR #2922, opened 2024-08-07
rebuilt tina lock file 

Calinator444:
### Description 

We're unable to edit the site using tina at the moment because the remote schema doesn't match the schema in the app.


amankumarrr:
closing this as there is 0 file changed with this PR and we need to clean up slot spaces.

PR #2960, opened 2024-08-15
💪 Fixing the TinaCMS experience

### LINKED ISSUE #2961: 💸 Remove DraftMode for TinaCMS
**Description**  

We were previously using DraftMode() from Next.js to integrate TinaCMS into our website, which significantly slowed down the entire TinaCMS experience in Tina Mode due to everything being rendered on the client side. This caused the website to become really slow, especially when switching to Tina Mode, as it would always enable DraftMode(). However, since the Tina team has now implemented app routing on their end, we can safely remove the draft mode from our website. This change should result in a much-improved user experience.

- [x] Remove the draftMode()
- [x] Adjust the existing page routing pages (i.e /products, /products/rewards)

amankumarrr:
Fixed #2961 

Affected Route:
- `/products`
- `/products/*`

**Screenshot:**

image
**Figure: Graphql became responsive after the change**


bradystroud (approved):
LGTM!
Everything seems to be much faster in TinaCMS ⚡

https://nextjs.org/docs/app/building-your-application/configuring/draft-mode
I was a bit confused about what draft mode did but Aman and Jack helped me understand.

PR #2966, opened 2024-08-17
🐛 BUG - Events dates are dynamic (different timezones), but time is static!

### LINKED ISSUE #1932: 🐛 BUG - Events dates are dynamic (different timezones), but time is static!
cc @camillars @adamcogan @Harry-Ross @amankumarrr @bradystroud 

**Describe the bug**

Looking at the dates for https://ssw.com.au/events/net-superpowers-tour

- From Australia we see 8th, 10th, and 12th which are correct for local time
- From Brazil we see 7th, 9th, and 11th meaning the dates are altered to users' timezone...

❌ This is bad because the times (9am-5pm) are NOT following the same behaviour - In other words... we are giving wrong information for people outside East Australia

**Expected behavior**

The date and time should be 100% clear by saying AEST - That is ideal for in-person events.

For online events it would be OK to show users' timezone as long as the time is also updated

- [x] 1. Please investigate and fix

--

Screenshot 2023-12-19 at 2 47 03 PM

See Eventbrite:

Screenshot 2023-12-19 at 2 52 49 PM

fenix2222:
- Affected routes: 

- Fixed #1932 

There is no standalone timepicker in tina. To keep existing data unchanged, I just added new end date picker which will be used for time.

new time picker component

**Figure**: **New time picker Component**
<br>

image
**Figure**: **Times using UTC**


components/training/eventBooking.tsx
```diff
@@ -96,7 +111,11 @@ const EventCard = ({ event, count, index, eventDurationInDays, schema }) => {
                 />
               </div>
               <div className="py-0.5 text-xs uppercase text-gray-500">
-                {EventModel.TIMINGS}
+                {event.date && event.endDate
+                  ? formatTimeWithAmPm(new Date(event.date)) +
+                    " - " +
+                    formatTimeWithAmPm(new Date(event.endDate))
+                  : EventModel.TIMINGS}
```
Calinator444: It can be a little confusing if you change the start time and the date doesn't change

```suggestion
                {
                event.date ? formatTimeWithAmPm(new Date(event.date) ) : "9AM" +
                " - " +
                event.endDate ? formatTimeWithAmPm(new Date(event.endDate)) : "5PM"
                }
```

amankumarrr (approved):
Looks good!

PR #2967, opened 2024-08-17
💸 Presenter has needlessly nested props

### LINKED ISSUE #2839: 💸 Presenter has needlessly nested props
@jeoffreyfischer @wicksipedia 

**Description**  
The `presenter` collection in tina has a nested `presenter` attribute that contains props that should be part of the main collection. This makes the collection confusing to work with because the presenter's name would be accessed from `presenter.presenter.name`

**Proposed solution**  

```[tasklist]
### Todo
- [ ] Update the collection so that `name` and `presenterProfileURL` are both part of the presenter collection
- [ ] Remove the `presenter` attribute from the `presenter` collection
- [ ] Update all of the existing presenters in tina to use this new schema]
- [ ] Update seeding on people repo to put presenter attributes in the correct place (see issue https://github.com/SSWConsulting/SSW.Website/pull/2916)
```

**Screenshot**

Image

fenix2222:
- Affected routes: 

- Fixed #2839


Calinator444:
@fenix2222 

@amankumarrr  is updating the cron job which syncs presenters from people. We would need to run it before merging this so that the Schema gets updated. Unfortunately I can't merge this right now because it would have breaking changes.


image
**Figure**: **Wicksy & Cooky's names missing due to schema change**

Calinator444:
Closing this PR 

We can't action this change unless we update the GitHub Action that auto fetches the SSW People profiles

PR #3065, opened 2024-09-04
2823 add missing filters events

### LINKED ISSUE #2823: 🐛Events - Past events not appearing in filter list
@wicksipedia 

### Describe the bug
The events page's counts do not include the aggregated past events in the filters until they've been loaded client side.

### Note
The fix probably involves doing a separate fetch to return the total number of events and using that number to populate the events filters.

### Tasks
- [ ] Investigate and fix it

### Screenshots

Image

**Figure**: **Note how the number of .Net Core events increases after more events load**

Calinator444:
### Affected Routes
- ssw.com.au/events

### Relevant issue
- Fixed #2823 

### Changes
- 🐛 Event filters will now show filters for the full list of events
- ✨ Clicking on a filter will load all events under that category
- ⚡ Once an event filter has been applied, events under that filter will be cached next time that filter is applied
- ♻️ Removed nested ternary operator used to handle loading state for events
- 🚸 Filtering for "Angular and React" events will now display Angular events
- 🧹 Tidied up events that were missing categories/had duplicate categories with different casing

### Screenshots

fetching functionality for events

**Figure**: **New fetching functionality for events that haven't been loaded**

<br>

event caching

**Figure**: **Caching behavior for filters that were previously applied**


components/filter/events.tsx
```diff
@@ -47,31 +51,42 @@ export type EventTrimmed = {
...
+  filterCategories,
   sidebarBody,
   defaultToPastTab,
 }: EventsFilterProps) => {
   const [pastSelected, setPastSelected] = useState<boolean>(defaultToPastTab);
+  const { past, upcoming } = filterCategories;
+  const { filters: futureFilters } = useEvents(upcoming);
+  const { filters: pastFilters } = useEvents(past);
+
+  useEffect(() => {}, [pastFilters, futureFilters]);
```
amankumarrr: do we need it there?

components/filter/events.tsx
```diff
@@ -140,87 +150,101 @@ const EventTab = ({ children }: { children: React.ReactNode }) => {
...
   We need this because there's an issue preventing us from syncing the files in the repo
   to Tina cloud. Images that aren't synced will 404.
    
    */
-  const [thumbnail, setFallbackImage] = useState(event.thumbnail);
+  const [oldTitle, setOldTitle] = useState(event.title);
+  const [visible, setVisible] = useState(false);
+  const [thumbnail, setFallbackImage] = useState("");
+  useEffect(() => {
+    if (event.title !== oldTitle) setVisible(false);
```
amankumarrr: can we put parenthesis around here? as per this rule - https://www.ssw.com.au/rules/how-to-avoid-problems-in-if-statements/#:~:text=See%20history.%20Try%20to

components/filter/events.tsx
```diff
@@ -140,87 +150,101 @@ const EventTab = ({ children }: { children: React.ReactNode }) => {
...
+    setTimeout(() => {
+      setVisible(true);
+      setOldTitle(event.title);
+    }, 100);
```
amankumarrr: can we avoid using timeout here?

amankumarrr (changes requested):
Looks good to me, just left a few comments, let me know if it doesn't make sense.

components/filter/events.tsx
Calinator444: Oh, my bad. I've deleted that now.

components/filter/events.tsx
Calinator444: Done - I've added parentheses around the if statement

components/filter/events.tsx
Calinator444: As per our conversation, I've improved the logic for hiding events so we no longer need a timeout

amankumarrr (approved):

PR #3068, opened 2024-09-05
🚑 Hotfix - Fix blank presenter on user group page

### LINKED ISSUE #2833: 💸 Update events endpoint to fetch presenters using the presenter list.
**Description**  
The presenter API on the website currently filters events by presenter name on the back end. This reduces performance because there are over 600 events meaning at worst the back end will filter through all 600+ events until it finds 10 events with the presenter name. We're doing the filtering here because some events with multiple presenters have the names all of the presenters in the presenterName field[.  Tina doesn't support fuzzy searching based on an attribute](https://tina.io/docs/graphql/queries/advanced/filter-documents/#operator-types) so we're doing the filtering on the back end for the time being.

**Proposed solution**

If we get rid of the `presenterName` field and move all of the presenters to the `presenterList` field we could do an equivalence check to search for events with a matching presenter instead of a fuzzy search, i.e. `event: {presenter: {name :{eq: "Adam Cogan"} }}`.

### Screenshot

Image

Calinator444:
### Description
Fixes a bug causing the presenter image to appear blank when not sourcing the presenter from Tina.

### Related Issue
- Fixed https://github.com/SSWConsulting/SSW.Website/issues/2833

### Screenshots

image

**Figure**: **Bad - presenter breaking with no presenter in list**

<br>

image

**Figure**: **Good - presenter sourcing name/link from event as a backup**

### DIFF
--- a/components/usergroup/organizer.tsx
+++ b/components/usergroup/organizer.tsx
@@ -1,4 +1,5 @@
 import Image from "next/image";
+import { useState } from "react";
 import type { Template } from "tinacms";
 import { tinaField } from "tinacms/dist/react";
 import type { TinaMarkdownContent } from "tinacms/dist/rich-text";
@@ -21,6 +22,7 @@ export const Organizer = ({
   data: OrganizerType;
   stringContent?: string;
 }) => {
+  const [image, setFallbackImage] = useState<string>(data.profileImg ?? "");
   return (
     <div className="flex flex-col gap-5">
       <div className="flex flex-row items-center gap-2">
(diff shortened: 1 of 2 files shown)


components/usergroup/organizer.tsx
```diff
@@ -28,13 +28,15 @@ export const Organizer = ({
...
-            height={68}
-            width={68}
-            className="w-17"
-          />
+          {data.profileImg && (
+            <Image
+              alt={`Picture of ${data?.name ?? "Organizer"}`}
+              src={data?.profileImg ?? ""}
+              height={68}
+              width={68}
```
amankumarrr: can we add a fallback for image if it returns 404? i.e onError prop for Image component

amankumarrr (changes requested):

components/usergroup/organizer.tsx
Calinator444: Done - I added a fallback so that if the image breaks it will be replaced with https://www.ssw.com.au/images/thumbs/avatar-thumbnail.png

amankumarrr (approved):

PR #3081, opened 2024-09-09
3073 add fallback presenter name

### LINKED ISSUE #3073: ♻️ Presenters - Add fallback behavior for presenter name field
### Description
Our events page uses the first presenter in the presenter list when rendering the event. However, this creates problems when the presenter list is not filled out (which will often happen if the presenter is external and not in Tina). 

```[tasklist]
- [x] Group the presenter list and presenter name field together in Tina
- [x] Update the hint for the presenter name field indicating that it can be used for external presenters
- [x] Update the SSW People profile link as it is not needed for internal presenters
- [x] use the the website to use presenterName as a backup if the first presenter is not provided
```

Calinator444:
- Affected routes: 

- Fixed #3073 

### Screenshots
image

**Figure**: **Testing a user group with no presenters**

<br>

image

**Figure**: **User Group tested with Adam added to the top of the presenter list**

<br>

separate people profile links for 2 preseenters

**Figure**: **Event has 2 presenters with individual profile links**

<br>

image

**Figure**: **Presenter label appears different when events have more than one presenter**


components/filter/events.tsx
```diff
@@ -350,25 +359,41 @@ const Event = ({ visible, event, jsonLd }: EventProps) => {
...
+              {(event.presenterName || event.presenterList?.length > 0) && (
                 <EventDescItem
-                  label="Presenter"
-                  linkValue={event?.presenterProfileUrl}
-                  value={event.presenterName}
-                />
+                  label={
+                    event?.presenterName || event?.presenterList?.length === 1
+                      ? "Presenter"
+                      : "Presenters"
```
amankumarrr, line 367: do you think else condition will ever executed on given scenario?

amankumarrr (changes requested):

components/filter/events.tsx
Calinator444: As per our conversation, the label should say "Presenter" so long as an overriding presenter name is provided.

amankumarrr (approved):

PR #3124, opened 2024-09-18
🗓️ Past Events - Fixed events endpoint

Calinator444:
CC: @ncn-ssw

### Description

The past events endpoint on the main site is failing due to an issue with Tina's cacheing system. I've disabled it on this branch to remedy the issue.


Calinator444:
Closing this - As per my conversation with @wicksipedia we would prefer to fix the cached not being written correctly to our docker image than disable it entirely for better performance.

PR #3128, opened 2024-09-19
🚧 Testing staging slot for exception on home page

Calinator444:
### Description

Created this PR to test if the runtime issues on the site are reproducible in staging


Calinator444:
Closing this - We've fixed the runtime error on the main site so we don't need this testing slot anymore

PR #3150, opened 2024-09-23
Removed unnknown unicode character in Nick's name

Calinator444:
### Description

There was a strange Unicode character in Nick's name. Removing the character so that this event still gets returned from the upcoming events endpoint.


amankumarrr (approved):

bradystroud (approved):
How did it get there? Was it added by Tina?

Calinator444:
@bradystroud 

> How did it get there? Was it added by Tina?

Based on the commit history this character was added when Andrew ran the script to pull events from Tina. I'm a little confused as to why there was a broken Unicode character at that point in Nick's name though. It makes no sense.

PR #3151, opened 2024-09-23
📈 App Router - added app insights telemetry for app components

### LINKED ISSUE #3153: ✨ App Routing - 📈Add telemetry for app router pages
CC: @wicksipedia 

### Description 

The App Router pages aren't sending Telemetry to app insights at the moment. This functionality was not added when the pages were ported. This needs to be fixed before rolling forward with the app router migration.

### Relevant routes

- /products
- /products/rewards

Calinator444:
### Description
This PR adds app insights telemetry to all existing App Router pages. Because the code has been added to `layout.tsx` the telemetry should be sent from all future pages to use app routing as well.

- Affected routes: 

- Fixed #3153


Calinator444:
**No longer relevant - I've made changes following these tests**

I have the requests being logged to app insights in staging

image

amankumarrr:
We should avoid using `<Suspense>` to wrap this component. Ideally, it should be a server-side component. Telemetry and App Insights should only wrap the child components. Perhaps we can try using `template.ts` for client-side rendering, which might allow us to send telemetry from the client-side.

amankumarrr:
maybe something like this:

```
"use client";
import { AppInsightsProvider } from "@/context/app-insight-client";
import { TelemetryProvider } from "./components/TelemetryProvider";

export default function Template({ children }: { children: React.ReactNode }) {
  return (
    <AppInsightsProvider>
      <TelemetryProvider>{children}</TelemetryProvider>
    </AppInsightsProvider>
  );
}

amankumarrr:
WebVitals are being logged correctly!
image

amankumarrr:
**Note**: The red arrow denotes the point in time when the page was migrated to app routing
Image

**Figure**: **Page views over time for /products**

amankumarrr:
Tested with upcoming UG, seems to be working fine!
image

Calinator444:
Worried about the missing `Suspense` boundary around the `LiveStream` component but Aman has confirmed with me that this issue does not result in errors when building the website locally.

Calinator444:
@wicksipedia @bradystroud 

I tested this to ensure it was sending requests to app insights in staging.

image

amankumarrr:
@bradystroud 
Fixed the issue for web vitals, as it was being called outside the AppInsight provider. 
image

bradystroud (approved):
LGTM - Aman and I checked out the results in app insights

PR #3232, opened 2024-10-18
Fixed sidebar going over the limits of the page

### LINKED ISSUE #3231: 🐛 About SSW - Left and right margins should be the same ⚠️
**Describe the bug**
as per @adamcogan @evecogan 

Image

It should be fixed by removing this extra class:

Image

**To Reproduce**
1. Visit https://www.ssw.com.au/company/about-us on iPad or reduce screen size

**Expected behavior**

Same padding on the right:

Image

tiagov8:
To fix the right spacing, I removed that extra class `md:block lg:min-w-96` from this DIV

Screenshot 2024-10-18 at 1 37 19 PM

@andrewwaltosssw 

I think this was caused by this change: https://github.com/SSWConsulting/SSW.Website/commit/87c3b2d7bc16ffc7e0feca0c7701789e1c43fd2b

1. Please double check my PR won't affect other pages you might have added that class for.

---

- Affected routes: https://www.ssw.com.au/company/about-us 

- Fixed https://github.com/SSWConsulting/SSW.Website/issues/3231

### DIFF
--- a/pages/company/[filename].tsx
+++ b/pages/company/[filename].tsx
@@ -83,7 +83,7 @@ export default function CompanyPage(
                 <div className="max-w-sm shrink pl-16">
                   {data.company.sidebar && (
                     <div
-                      className={classNames("md:block lg:min-w-96", {
+                      className={classNames("md:block", {
                         hidden: data.company.hideSidebarOnMobile,
                         "min-w-fit": data.company.fixedWidthSidebar,
                       })}


andrewwaltosssw (approved):
Nice pickup. This change seems reasonable. Please check to make sure layout still makes sense on mobile resolutions when the elements get rearranged.

PR #3242, opened 2024-10-22
Upgrading app service plan to P1V3

### LINKED ISSUE #3243: 🐛 Website crashes when there are too many PR slots
@amankumarrr 
@wicksipedia 

**Describe the bug**
When there are too many PR slots, the app service plan CPU and RAM gets maxed out - and the website crashes.

**To Reproduce**
1. Open lots of PRs (14 were open last time this happened)

**Expected behavior**
We should be able to have a reasonable number of slots running without the site crashing

As per my email: **SSW.Website - investigate scaling options**

**plan-sswwebsite** is currently running on the **P1v2** SKU.
We could set up some automated scale up/out rules, but a better option is just to upgrade the SKU to **P1v3**.
With a 3-year reservation, we will end up paying less per month.

SKU	        CPU cores	RAM (GB)	$/month	$/month - 3yr res.	
P1v2	1	                3.5	                130	        N/A	    < - current (A$130/month)
P1v3	2	                8	                195	        88	    < - recommended (A$88/month)
More info here: https://azure.microsoft.com/en-us/pricing/details/app-service/linux/

Annual pricing will change from **A$1,560** to **A$1,056**.

So, I plan to:
1.	Change the SKU to P1v3 in bicep and deploy
2.	Create a 3-year reservation: https://portal.azure.com/#view/Microsoft_Azure_Reservations/ReservationsBrowseBlade/productType/Reservations
3.	With the SSW.Website team, keep an eye on CPU/RAM usage over the next 2 weeks
        - If we need more resources still, we’ll split prod/non-prod to separate service plans

chrisschultzssw:
Upgrading the app service plan to P1V3 - giving us more CPU & RAM.
SKU details here: https://azure.microsoft.com/en-us/pricing/details/app-service/linux/

- Fixed: #3243 
i.e. the website was crashing when there were too many PR slots running.

We also found that the SKU was being set twice - we removed this from appService.bicep and kept it in main.bicep.

### DIFF
--- a/infra/appService.bicep
+++ b/infra/appService.bicep
@@ -5,30 +5,11 @@ param tags object
 param dockerRegistryServerURL string
 param appInsightConnectionString string
 param keyVaultName string
+param skuName string
+param skuCapacity int
 
 param healthCheckPath string = '/'
 
-@allowed([
-  'B1'
-  'B2'
-  'B3'
-  'S1'
-  'S2'
-  'S3'
-  'P1'
-  'P2'
-  'P3'
-  'P1V2'
-  'P2V2'
-  'P3V2'
-  'P1V3'
-  'P2V3'
-  'P3V3'
-])
-param skuName string = 'P1V2'
-
-@minValue(1)
-param skuCapacity int = 1
 
 param acrName string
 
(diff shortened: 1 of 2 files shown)


amankumarrr (approved):
LGTM

PR #3348, opened 2024-11-20
Update logo index

bettybondoc:
cc @hveraus @tiagov8 

- Affected routes: /logo/ssw
- PBI: [Product Backlog Item 97260](https://ssw.visualstudio.com/SSW.Design/_workitems/edit/97260): 💄 SSW.Design - Change official webfont from Open Sans to Inter

Upload new updated SSW logos with taglines using Inter


hveraus (changes requested):
Hi Betty. 

The new images suppose to be exactly the same size as the old one. 

But I think the large PNG should be OK. The new preview image is 360x220 which is different from the original 360x216, might cause some pixel perfect problem. better safe to change it back to 360x216px

bettybondoc:
> Hi Betty.
> 
> The new images suppose to be exactly the same size as the old one.
> 
> But I think the large PNG should be OK. The new preview image is 360x220 which is different from the original 360x216, might cause some pixel perfect problem. better safe to change it back to 360x216px

@hveraus - done :)

hveraus (approved):

PR #3395, opened 2024-12-11
New banner for the .net superpowers

PennyWalker:
New banner for the .net superpowers - Newsletter


vladislav-kir (approved):
Looks good to me :)

PR #3403, opened 2024-12-13
Update and rename _2024_12__Adoption % – Have Developers Embraced Git…

PennyWalker:
Changing the name of the file as well as the URL


SamBWagner (approved):

wicksipedia:
used admin rights to bypass required check - only a static html file was modified

PR #3404, opened 2024-12-13
Fixing linting errors and update packages

### LINKED ISSUE #3382: 🐛 Eslint breaking our builds
### Description

Next.js is performing it's own listing check for the website at build time which does not align with the linting check in our pipeline. As a result we cannot merge any PRs right now because both linting checks have separate criteria.

Todo:
- [ ] Update the linting check for the website to align with Next.js' linting check

Image

babakamyljanovssw:
- Fixes #3382

Changes:

- Turned off this rule `tailwindcss/classnames-order` - because it was suggesting to replace the order of classes with incorrect ordering as per [the docs](https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted). Even changing the order as per the suggestion and then saving the file was automatically reverted the changes to previous state. Tried to update the plugin but didn't work. 
- To resolve the error `tailwindcss/no-custom-classname` for this class`prose-h2:text-3xl/9`, I've used [Javascript classNames utility](https://github.com/JedWatson/classnames) to make it consistent with other occurrences of that class that were using utility to avoid linting issue.
-  Added new regex `prose-.+` to `.eslintrc` to ignore `prose-dark` and `prose-oppurtunity` classes.
- Updated dev dependency packages.


.eslintrc
```diff
@@ -21,10 +21,11 @@
     "quotes": ["error", "double"],
     "no-console": "warn",
     "tailwindcss/no-arbitrary-value": "error",
+    "tailwindcss/classnames-order": "off",
     "tailwindcss/no-custom-classname": [
       "error",
       {
-        "whitelist": ["g\\-yt.+", "legend", "dark"],
+        "whitelist": ["g\\-yt.+", "prose-.+", "legend", "dark"],
       },
```
Calinator444: I'd add text-3xl/9 to the white list explicitly instead of using "classNames"


```suggestion
        "whitelist": ["g\\-yt.+", "prose-.+", "legend", "text-3xl/9","dark"],
      },
```

components/company/companyPageCard.tsx
```diff
@@ -42,10 +43,11 @@ const CompanyPageCard = ({ data, schema, index }) => {
...
 };
 
+const proseClass = "prose-h2:text-3xl/9";
 const CompanyPageCardContent = ({ title, body, schema, index }) => {
   return (
     <article className="col-span-1 size-full rounded border-1 border-gray-300 bg-white px-8 py-4 shadow hover:border-ssw-black dark:border-gray-700 dark:bg-gray-800">
-      <div className="prose prose-h2:text-3xl/9">
+      <div className={classNames("prose", proseClass)}>
         <h2
           className="my-1"
```
Calinator444: Just leave this the way it was, it's better to use the whitelist to keep track of which classes are exempt from the linting check

```suggestion
      <div className="prose prose-h2:text-3xl/9">
        <h2
          className="my-1"
```

package.json
```diff
@@ -32,12 +32,12 @@
     "@types/js-cookie": "^3.0.6",
     "@types/node": "^20.17.6",
     "@types/react": "^18.3.12",
-    "@typescript-eslint/eslint-plugin": "^8.17.0",
-    "@typescript-eslint/parser": "^8.17.0",
+    "@typescript-eslint/eslint-plugin": "^8.18.0",
+    "@typescript-eslint/parser": "^8.18.0",
```
Calinator444, line 36: There's no need to update these. The caret syntax will automatically resolve these dependencies to the latest major and  minor versions.

No need to update this either, just food for thought

https://www.geeksforgeeks.org/difference-between-tild

.eslintrc
babakamyljanovssw: Done

components/company/companyPageCard.tsx
babakamyljanovssw: Done

Calinator444 (approved):

PR #3406, opened 2024-12-13
Update _2024_12__Adoption Percentage – Have Developers Embraced GitHu…

PennyWalker:
fixing x2 incorrect URLs


StefStar (approved):

wicksipedia:
Used admin privileges - only static html was updated

PR #3458, opened 2025-01-07
💄 Consultingv2 – Setting Card Index to Center on Tab Change / Initial Render

### LINKED ISSUE #3442: ✨ Cards Component – Default Card Positions
**Is your feature request related to a problem? Please describe.**
Choosing a card tab looks unnatural if it starts at the end or start.

**Describe the solution you'd like**
There's an extra prop going into the card components that sets the center card to a certain index.

**Additional context**
As per the 3/1/2025 sprint review, feedback from @louisroa8189.

isaaclombardssw:
- Affected routes: consulting v2 components.
- For card-based components, in the carousel, sets the index on initial render and tab change to the centre index (or lower of centre indices for even length card lists).

- Fixed #3442.

### DIFF
--- a/components/ui/carousel.tsx
+++ b/components/ui/carousel.tsx
@@ -18,6 +18,7 @@ type CarouselProps = {
   plugins?: CarouselPlugin;
   orientation?: "horizontal" | "vertical";
   setApi?: (api: CarouselApi) => void;
+  itemLength?: number;
 };
 
 type CarouselContextProps = {
@@ -54,6 +55,7 @@ const Carousel = React.forwardRef<
       plugins,
       className,
       children,
+      itemLength,
       ...props
     },
     ref
@@ -69,7 +71,7 @@ const Carousel = React.forwardRef<
         AutoPlay({
           delay: 5000,
           playOnInit: true,
-          stopOnInteraction: false,
+          stopOnInteraction: true,
           stopOnMouseEnter: true,
         }),
         ...(plugins || []),
@@ -135,6 +137,11 @@ const Carousel = React.forwardRef<
         api?.off("select", onSelect);
       };
     }, [api, onSelect]);
+
+    React.useEffect(() => {
+      api?.scrollTo(Math.floor((itemLength - 1) / 2));
+      setSelectedIndex(Math.floor((itemLength - 1) / 2));
+    }, [itemLength, api]);
 
     return (
       <CarouselContext.Provider
--- a/components/ui/carousel.tsx
+++ b/components/ui/carousel.tsx
@@ -139,6 +139,9 @@ const Carousel = React.forwardRef<
     }, [api, onSelect]);
 
     React.useEffect(() => {
+      if (!api || !itemLength) {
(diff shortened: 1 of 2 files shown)


components/ui/carousel.tsx
```diff
@@ -69,7 +71,7 @@ const Carousel = React.forwardRef<
         AutoPlay({
           delay: 5000,
           playOnInit: true,
-          stopOnInteraction: false,
+          stopOnInteraction: true,
           stopOnMouseEnter: true,
```
Calinator444, line 75: Why was this enabled? If `stopOnInteraction` is enabled the card carousel will stop as soon as somebody drags it.

components/ui/carousel.tsx
```diff
@@ -136,6 +138,11 @@ const Carousel = React.forwardRef<
       };
     }, [api, onSelect]);
 
+    React.useEffect(() => {
+      api?.scrollTo(Math.floor((itemLength - 1) / 2));
+      setSelectedIndex(Math.floor((itemLength - 1) / 2));
+    }, [itemLength, api]);
+
```
Calinator444: if index.length has an initial value of "0" couldn't this result in the index being out of range?

image

isaaclombardssw: Yep – I've tested it with new groups, etc. and it doesn't cause any problems with an empty list. But so it's a bit cleaner I'll bypass that code if itemLength is 0 :)

components/ui/carousel.tsx
isaaclombardssw: As per our conversation, we'll just leave this one on too, so users get to read cards.

components/ui/carousel.tsx
isaaclombardssw: Pushed-

components/ui/carousel.tsx
```diff
@@ -136,6 +138,14 @@ const Carousel = React.forwardRef<
...
     }, [api, onSelect]);
 
+    React.useEffect(() => {
+      if (!api || !itemLength) {
+        return;
+      }
+      api?.scrollTo(Math.floor((itemLength - 1) / 2));
+      setSelectedIndex(Math.floor((itemLength - 1) / 2));
+    }, [itemLength, api]);
+
```
Calinator444: returning inside of a useEffect doesn't exit the function (assuming that's what you're trying to do). It's used for adding cleanup code. See [docs](https://react.dev/reference/react/useEffect#connecting-to-an-external-system)

```suggestion

    React.useEffect(() => {
      const lengthOrApiMissing = !api || !itemLength
      if (!lengthOrApiMissing ) {
           api?.scrollTo(Math.floor((itemLength - 1) / 2));
           setSelectedIndex(Math.floor((itemLength - 1) / 2));
      }
    }, [itemLength, api]);

```

isaaclombardssw: Forgot about that one – good catch

Calinator444 (approved):

PR #3509, opened 2025-01-16
💄Footer - Add Azure/TinaCMS banner to all page templates

### LINKED ISSUE #3498: Footer - missing from homepage + needs more info
> Hi @wicksipedia , 
> 
> as per my message on Teams... the section "BUILT ON / POWERED BY" is not currently showing
> 
> - [ ] 1. Can you implement it back?
> 
> Image

>  ~~Also, to make SSW.Website consistent with Rules (https://github.com/SSWConsulting/SSW.Rules/pull/1576) and People (TODO https://github.com/SSWConsulting/SSW.People/issues/663)...~~
> 
> ~~[ ] 2. Add after the logos, the sentence "**We ❤️ open source. Loving SSW Website? [Star us on GitHub.](https://github.com/SSWConsulting/SSW.Website)**" + the Star count~~
> 

 _Originally posted by @tiagov8 in [#2711](https://github.com/SSWConsulting/SSW.Website/issues/2711#issuecomment-2590971735)_

Calinator444:
- Affected routes: /*

### Description 
This PR adds the azure banner to every page. There are page templates where the default color of the banner doesn't fit the style of the page so I've updated the banner to default to a color that suits the content. I've also added an option in Tina to disable the banner where it might not suit the content of the page. 

**Note**: **"white" has never been a valid property value for the section color property. I just added typescript to this prop so the errors are appearing now.**

### Changes
-  Added a the Azure banner to every page without one
- ♻️ Created a more efficient way to add fields to every page template
- ✨ Added an option to change the background of the Azure banner on every page
- ✨ Added an option to disable the azure banner on every page

- Fixed #3498

azure banner customizeability

**Figure**: **Customizing the Azure banner**

### DIFF
--- a/app/articles/[filename]/index.tsx
+++ b/app/articles/[filename]/index.tsx
@@ -138,7 +138,7 @@ const ArticlePage = ({ props, tinaProps }: ArticlePageProps) => {
             ? data.articles
             : {
                 azureBanner: {
-                  azureFooterColor: "default",
+                  azureFooterColor: "white",
                 },
               }
         }
--- a/app/articles/[filename]/index.tsx
+++ b/app/articles/[filename]/index.tsx
@@ -79,7 +79,7 @@ const ArticlePage = ({ props, tinaProps }: ArticlePageProps) => {
       {data.articles.subTitle && (
         <section
           className={classNames(
-            "prose mx-auto w-full max-w-9xl flex-row px-8 pb-8 prose-h1:my-0 prose-h1:pt-8 prose-h2:mt-8 prose-img:my-0 lg:flex"
+            "prose prose-h1:my-0 prose-h1:pt-8 prose-h2:mt-8 prose-img:my-0 mx-auto w-full max-w-9xl flex-row px-8 pb-8 lg:flex"
           )}
         >
           <div data-tina-field={tinaField(data.articles, "subTitle")}>
--- a/app/articles/[filename]/index.tsx
+++ b/app/articles/[filename]/index.tsx
@@ -1,6 +1,7 @@
 "use client";
 
 import ArticleAuthor from "@/components/articles/articleAuthor";
+import { BuiltOnAzure } from "@/components/blocks/builtOnAzure";
 import { componentRenderer } from "@/components/blocks/mdxComponentRenderer";
 import { CallToAction } from "@/components/callToAction/callToAction";
 import { PreFooter } from "@/components/layout/footer/pre-footer";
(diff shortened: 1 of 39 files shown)


app/articles/[filename]/index.tsx
```diff
@@ -25,10 +23,6 @@ export type ArticlePageProps = {
 
 const ArticlePage = ({ props, tinaProps }: ArticlePageProps) => {
   const { data } = tinaProps;
-  const showCallToAction = data.articles.callToAction.showCallToAction;
```
Comment (isaaclombardssw) [inline app/articles/[filename]/index.tsx:28]: I don't think you want to remove this! 🫨

app/consulting/[filename]/consulting2.tsx
```diff
@@ -24,7 +23,7 @@ const Consulting2 = memo(
     return (
       <>
         <div className="dark flex h-full flex-col">
-          <Section color={SectionColor.ToggleLightMode}>
+          <Section color={"toggleLightMode"}>
```
Comment (isaaclombardssw) [inline app/consulting/[filename]/consulting2.tsx:26]: Not sure this is necessary, but nbd.

app/industry/index.tsx
```diff
@@ -25,6 +25,7 @@ export default function IndustriesPage({ tinaProps }) {
           </div>
         </div>
       </Container>
+      <BuiltOnAzure data={data.industryIndex} />
```
isaaclombardssw: I think this looks a little confusing, like it's passing in totally different data for every page

ex. 

\<BuiltOnAzure data={data.industryIndex} />

vs 

\<BuiltOnAzure data={data.offices} />

My first thought is why does one need an industry index and one need offices??

components/layout/layout.tsx
```diff
@@ -155,8 +152,6 @@ export const Layout = ({
           </header>
           <ErrorBoundary key={router.asPath}>
             <main className={classNames("grow bg-white")}>{children}</main>
-
-            {showAzureBanner && <PreFooter />}
```
isaaclombardssw, line 159: This approach is so much easier to maintain. I can see why you've chucked it in the child classes instead (but then we need to add it into each).

It's a bit annoying – maybe add something to the readme to tell people to add it (your reason for having it like this does make sense). OR find some workaround.

isaaclombardssw (changes requested):
Really good work 🚀 

Just a couple of suggested improvements

app/articles/[filename]/index.tsx
Comment (Calinator444) [reply app/articles/[filename]/index.tsx:28]: It's unused anyway.

app/industry/index.tsx
Calinator444: as per our conversation I've updated the footer template to accept <page>.azureBanner as an argument instead for readability

isaaclombardssw (approved):
🚀  very nice

PR #3687, opened 2025-03-06
Adding Lighthouse to audit the pages

### LINKED ISSUE #3675: 💸 Pipeline - Evaluate the performance of important pages
**Description**
Currently, we are only evaluating the performance of the homepage using PagespeedInsights, and there is significant variation in results each time.

LightHouse should check important pages: Home, Consulting...

- [x] Change action to use LightHouse
- [x] Add summary for the main build
- [x] Create rule

**Proposed Solution**
As discussed with @wicksipedia, we should analyze all key pages using Lighthouse to ensure consistency:

- `/` ⭐
- `/articles`
- `/company/about-us`
- `/company/clients`
- `/company/contact-us`
- `/consulting`
- `/consulting/net-upgrade` ⭐
- `/consulting/web-applications` ⭐
- `/employment`
- `/events/angular-superpowers-tour`
- `/events/ai-workshop`
- `/netug/sydney`
- `/training/internship-fullstack`

amankumarrr:
- Affected routes: 

- Fixed #3675 

- Adding the Lighthouse report for the PR

It adds up additional 5+ minutes for the PR checks


scripts/generate-lighthouse-report.py
```diff
@@ -0,0 +1,99 @@
...
+        accessibility = result["summary"]["accessibility"] * 100
+        best_practices = result["summary"]["best-practices"] * 100
+        seo = result["summary"]["seo"] * 100
+
+        total_bundle_size, unused_bundle_size = get_total_and_unused_bytes_for_url(url)
+        parsed_url = urlparse(url)
+        url_display = f"⭐ {url}" if parsed_url.path in important_paths else url
+
+        mdx_output.append(
+            f"| {url_display} | {performance:.2f} | {accessibility:.2f} | {best_practices:.2f} | {seo:.2f} | {total_bundle_size:.2f} MB | {unused_bundle_size:.2f} MB |"
```
wicksipedia: render the numbers as integers not floats

wicksipedia:
bundle size & unused bundle size - what do they mean in the context of a user opening that url?

amankumarrr:
> bundle size & unused bundle size - what do they mean in the context of a user opening that url?

@wicksipedia When the user loads the page, they are aware of the bundle sizes being loaded, which could cause delays. However, it is more meaningful for developers to see if their pull requests are introducing additional bundles that might impact the page's performance.

scripts/generate-lighthouse-report.py
amankumarrr: done

wicksipedia:
> > bundle size & unused bundle size - what do they mean in the context of a user opening that url?
> 
> @wicksipedia When the user loads the page, they are aware of the bundle sizes being loaded, which could cause delays. However, it is more meaningful for developers to see if their pull requests are introducing additional bundles that might impact the page's performance.

Makes sense but from the PR comment, I can't tell if the bundle size is changing.

I think it would be useful if there was an extra emoji that showed if the bundle size was smaller/larger than `main`. This can be in a separate PBI.

.github/workflows/pr-push-deploy.yml
```diff
@@ -102,40 +102,52 @@ jobs:
...
-      - uses: wicksipedia/pagespeed-insights@main
-        id: pagespeed-insights
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+      - name: Wait for Slot to be available
+        uses: jtalk/url-health-check-action@v2
+        with:
+          url: https://www.ssw.com.au
```
wicksipedia, line 119: should this reference the slot url?

wicksipedia:
```suggestion
          url: ${{ needs.pr-deploy.outputs.url }}
```

wicksipedia (approved):

PR #3707, opened 2025-03-12
Fixing the issue with dynamically created pages

### LINKED ISSUE #2643: 🐛 Unable to edit newly created pages in Tina 🐛
cc: @amankumarrr @wicksipedia @samsmithssw @scottgallant
**Describe the bug**
When creating a new page via Tina Admin on production site, I am unable to edit the page due to getting a 500 and/or 404 error

**To Reproduce**
1. Go to www.ssw.com.au/admin and login
2. Create a new branch
3. Tina Menu | Consulting - Pages
4. Click "Create New +"
5. Add a filename and click "Save"
6. After auto-reload find the newly created file and click it
7. Preview shows 500 error and sometimes 404 error

**Additional Info**
This behaviour does not happen when adding a page from Tina Admin on a PR slot URL

**Acceptance Criteria**
- see the newly created file to make my content edits
- get the fix on TinaCloud starter

```[tasklist]
### Tasks
- [x] ~Add a draft mode OR ISR setting to pages that allows visual editing for newly added pages (see what works).~ Enabled dynamic Routing
- [ ] If successful, update the Tina.io documentation to include a tutorial on how to achieve this effect (I believe it would be a common complaint with the editorial workflow).
- [ ] If successful, notify the Tina.io team of the workaround (they have had a similar problem) and work with them to set up the fix.
```

Note: If this suggested solution doesn't work, work with Jack P to find a fix.

**Screenshots/Video**
Image
**Figure: The error i'm seeing**

amankumarrr:
Fixed: #2643

- Removing the `dynamicParam` to handle dynamic routes for newly created Tina pages.
- Adding guards to handle the null errors with minimum props


app/[filename]/page.tsx
```diff
@@ -47,7 +45,7 @@ export async function generateMetadata({
   }
 
   // eslint-disable-next-line react-hooks/rules-of-hooks
-  const { seoProps } = useSEO(seo);
+  const { seoProps } = useSEO(seo || {});
```
Comment (Calinator444) [inline app/[filename]/page.tsx]: Could we do something more like this please? remembering to add `|| {}` everywhere will be difficult

image

Calinator444 (changes requested):

app/[filename]/page.tsx
Comment (amankumarrr) [reply app/[filename]/page.tsx]: Yeah good idea, just added it.

Calinator444 (approved):

PR #3776, opened 2025-03-24
Upgrading headlessUI to 2.2.0

### LINKED ISSUE #2683: 💸 @headlessui/react from 1.7.19 to 2.0.3
**Description**  
We are using carousels from the @headlessui/react library, but the latest version is not compatible with our implementation.

**Proposed solution**  
Ideally, our implementation should be aligned with the latest versions of the library.

- [x] Investigate and work with the latest version.
- [x] Update it for ssw.megamenu library as well

Dependabot requested upgrade with this PR - https://github.com/SSWConsulting/SSW.Website/pull/2625

amankumarrr:
Affected Route: `/*`
Fixed #2683


components/filter/FilterGroup.tsx
```diff
@@ -75,6 +75,7 @@ const FilterOption = ({
     >
       <div className="inline-block h-3.5 w-6 text-sswRed">
         <Transition
+          as={"div"}
```
babakamyljanovssw: to make consistent with others removed curly brackets
```suggestion
          as="div"
```

babakamyljanovssw (approved):

PR #4140, opened 2025-08-05
Create notify-n8n-of-articles.yml

sethdaily:
As per [Product Backlog Item 114344](https://dev.azure.com/ssw/SSW.Marketing/_workitems/edit/114344): 🤖 Social Media - new trigger - articles

Adding a webhook action to send new articles to n8n. They will be used in n8n to create social media posts for sharing on our socials.

### DIFF
--- a/.github/workflows/notify-n8n-of-articles.yml
+++ b/.github/workflows/notify-n8n-of-articles.yml
@@ -1,0 +1,48 @@
+name: Notify n8n of new articles
+
+on:
+  push:
+    branches: [main]
+    paths:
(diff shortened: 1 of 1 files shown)


wicksipedia (approved):
This would be a good use case for TinaCloud's webhook support but I suspect that you are doing it this way since you are sending the  content of the file to n8n as well

sethdaily:
@wicksipedia 
> This would be a good use case for TinaCloud's webhook support but I suspect that you are doing it this way since you are sending the content of the file to n8n as well

Yes that's correct

PR #4206, opened 2025-09-16
Update _2025_09__MCP as Your Integration Layer_.html

PennyWalker:
updating typo


jernejk (approved):
The URL seems to be ok now.

PR #4403, opened 2025-12-16
updating all SEO titles for consulting pages

uly1:
I've used cursor to update all 80+ consulting page titles. I'll sit with @StefStar to make sure we agree they're good


StefStar (approved):
LGTM

PR #4719, opened 2026-05-29
consultingv2/openclaw (PR from TinaCMS)

StefStar:


Levijj22 (changes requested):
Thinking change "consultation" to "engagement" so they know we will actually do something, not just consult.

Levijj22 (approved):

PR #4724, opened 2026-06-03
add mobile breadcrumb

micsblank:
Making breadcrumbs mobile reactive and rule compliant


app/components/breadcrumb.tsx
```diff
@@ -118,14 +118,50 @@ export const Breadcrumbs: FC<BreadcrumbsProps> = ({
...
+          {breadcrumbItems}
+        </BreadcrumbList>
+      </Breadcrumb>
+      <nav className="sm:hidden" aria-label={`Back to ${mobileParent.label}`}>
+        <a
+          href={mobileParent.href}
+          className="unstyled inline-flex items-center gap-1 text-xs text-gray-700 no-underline hover:text-sswRed"
+          aria-label={`Back to ${mobileParent.label}`}
+        >
+          <svg
```
joshbermanssw: Instead of <svg> can we use a component from the component library (whatever the chevron is from the other breadcrubm list)

components/blocks/breadcrumbs/breadcrumbs.tsx
```diff
@@ -121,6 +134,32 @@ export function Breadcrumbs({ data }: { data: Consultingv2BlocksBreadcrumbs }) {
...
+        <nav
+          className="text-gray-300 sm:hidden"
+          aria-label={`Back to ${mobileParent.label}`}
+        >
+          <a
+            href={mobileParent.href}
+            className="unstyled inline-flex items-center gap-1 text-sm no-underline transition-colors hover:text-white"
+            aria-label={`Back to ${mobileParent.label}`}
+          >
+            <svg
```
joshbermanssw: same here - no need to use sg

joshbermanssw (changes requested):

app/components/breadcrumb.tsx
micsblank: @copilot

components/blocks/breadcrumbs/breadcrumbs.tsx
micsblank: @copilot  same here, no need to use SVG, please use an existing component

components/blocks/breadcrumbs/breadcrumbs.tsx
```diff
@@ -103,10 +103,23 @@ export function Breadcrumbs({ data }: { data: Consultingv2BlocksBreadcrumbs }) {
...
+  const mobileParent =
+    segments.length >= 2
+      ? {
+          label:
+            global.breadcrumbReplacements.find(
+              (value) => value.from === segments[segments.length - 2]
+            )?.to || segments[segments.length - 2],
+          href: `/${segments[segments.length - 2]}`,
+        }
+      : { label: global.breadcrumbHomeRoute, href: "/" };
```
joshbermanssw, line 116: This `.find(...)?.to || rawSlug` lookup is a bit convoluted, and it's a hand-written second copy of the same slug→label logic that lives in `app/components/breadcrumb.tsx` (`getDisplayName`). Consider a small `Map` instead, e.g. build it once:

```ts
const replacementMap = new Map(
  global.breadcrumbReplacements.map((r) => [r.from, r.to])
);
const parent = segments[segments.length - 2];
const label = replacementMap.get(parent) ?? parent;
```

Cleaner to read, and better still if the whole mobile back-link were extracted into one shared component so the two breadcrumb files can't drift (the `href` already differs between them — full path here vs. single segment).

components/blocks/breadcrumbs/breadcrumbs.tsx
```diff
@@ -121,6 +134,32 @@ export function Breadcrumbs({ data }: { data: Consultingv2BlocksBreadcrumbs }) {
...
+              height="12"
+              viewBox="0 0 24 24"
+              fill="none"
+              stroke="currentColor"
+              strokeWidth="2"
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              aria-hidden="true"
+            >
+              <path d="M15 18l-6-6 6-6" />
```
isaaclombardssw: Extract this out

isaaclombardssw (approved):
Please refactor out this svg (or use a real icon component), otherwise it's okay

joshbermanssw (approved):

PR #4726, opened 2026-06-03
Improve Lighthouse performance & accessibility on event/consulting pages

joshbermanssw:
## TLDR;
Performance and accessibility fixes from a Lighthouse audit of `/events/ai-workshop-essentials` — trims render-blocking CSS, fixes the LCP image, stops the `/events/null` request, and labels the carousel controls.

## Changes:
- [ ] Replace `classNames` (from `tinacms`) with the local `cn` helper to keep the CMS package out of the client bundle
- [ ] Add `priority` + `sizes` to the block image and drop `loading="eager"` to improve LCP image delivery
- [ ] Guard the events catch-all route with `notFound()` so a `null` slug stops firing `/events/null` + dead Tina queries
- [ ] Enable `experimental.inlineCss` to remove render-blocking CSS requests
- [ ] Add `aria-label`, `aria-current` and `type="button"` to carousel dot buttons (accessible name)
- [ ] Remove `text-decoration:none` from footer deployment links so they're distinguishable


designbyalex:
@joshbermanssw 

57% on PageSpeed Insights

https://pagespeed.web.dev/analysis/https-app-sswwebsite-9eb3-pr-4726-azurewebsites-net-events-ai-workshop-essentials/3vj9vugj5b?form_factor=mobile

components/blocks/imageComponents/ImageComponentLayout.tsx
```diff
@@ -86,9 +86,10 @@ export const ImageComponentLayout = ({ data, children }) => {
               isImage && (
                 <Image
                   width={data.mediaConfiguration?.imageWidth}
-                  loading="eager"
+                  priority
```
kulesy: **Major — `priority` is unconditional, so every instance preloads as if it were the LCP**

`ImageComponentLayout` is used by `imageTextBlock` and `accordionBlock`, both of which are Tina blocks an editor can place multiple times per page. Verified locally against `/events/ai-workshop-essentials` (4 `imageTextBlock` instances): the rendered HTML emits a `<link rel="preload" as="image">` for each one, on top of the actual hero/LCP preload and the header logo. That's 4 content-image preloads competing for the LCP slot when only one of them is the LCP candidate.

Chrome warns that `priority` should be reserved for the one LCP image; multiple preloads dilute the browser's prioritization signal and can hurt LCP rather than help it.

Suggested fix: expose `priority` as a prop on `ImageComponentLayout` (default `false`) and opt in only on the consumer that needs it (the first block above-the-fold). Or derive it from "this is the first block of the page".

components/videoModal.tsx
```diff
@@ -112,51 +108,45 @@ export const VideoModal = ({
...
           <div className="size-full" onClick={() => setClicked(true)}>
-            {videoState.imageSrc && (
+            {imageSrc && (
               <>
                 <Image
                   className={classNames("!my-0", frameClassName)}
-                  src={videoState.imageSrc}
+                  src={imageSrc}
                   fill
+                  priority
```
kulesy: **Major — every `VideoModal` on the page now preloads its thumbnail**

Same issue as the image-block one: `priority` is unconditional. Verified locally on `/events/ai-workshop-essentials`, which has 4 VideoModals — each emits a `<link rel="preload" as="image" imageSrcSet="...maxresdefault.jpg...">` for its YouTube thumbnail.

Before this PR, `Image` here was `dynamic(..., { ssr: false })` and lazily loaded, so the thumbnails didn't compete for LCP attention. Going server-rendered + `priority` reverses that — below-the-fold YouTube thumbnails now race the hero image for bandwidth on slow connections.

Suggested fix: accept a `priority?: boolean` prop on `VideoModal` (default `false`), and pass it through to `<Image>`. The consumer that owns the actual LCP video can opt in.

app/(events)/events/[...filename]/page.tsx
```diff
@@ -210,6 +211,10 @@ export default async function Events(prop: {
     getData(slug.join("/")),
   ]);
 
+  if (!filename || filename === "null") {
+    notFound();
+  }
```
Comment (kulesy) [inline app/(events)/events/[...filename]/page.tsx]: **Major — `notFound()` runs *after* the Tina queries, so the "dead Tina queries" claim doesn't hold**

Verified locally: requesting `/events/null` produces these in the dev log *before* `notFound()` fires:

```
Unable to find record content/eventsv2/null.json    ← newEventsPageData ran
Unable to find record content/events/null.mdx       ← getData ran
```

The user-visible 404 is achieved, but the wasteful Tina round-trips still happen on every null-slug request — which is the part the PR description says this fix addresses.

Move the guard above the `await Promise.all([...])`:

```ts
const slug = params.filename;
const filename = slug[slug.length - 1];

if (!filename || filename === "null") {
  notFound();
}

const [newPage, calendarEvent, oldPage] = await Promise.all([...]);
```

app/consulting/[filename]/page.tsx
```diff
@@ -178,6 +179,8 @@ export default async function Consulting(props: {
     consultingPageData(filename),
   ]);
 
+  if (!filename || filename === "null") notFound();
```
Comment (kulesy) [inline app/consulting/[filename]/page.tsx]: **Major — same as the events-page guard: runs after the Tina fetch**

Same fix as the [events-page comment](#) — move the `notFound()` check above `await Promise.all([newConsultingPageData(filename), consultingPageData(filename)])` so the dead Tina queries are actually short-circuited, not just the render.

components/blocks/imageComponents/ImageComponentLayout.tsx
joshbermanssw: Fixed in 41ba1532. `ImageComponentLayout` now takes a `priority` prop (default `false`) and `blocks-renderer` only sets `priority: i === 0`, so just the first block opts in as the LCP candidate instead of every image-block instance preloading.

components/videoModal.tsx
joshbermanssw: Fixed in 41ba1532. `VideoModal` now accepts `priority?: boolean` (default `false`) and passes it through to `<Image>`; only the first block opts in via `blocks-renderer`, so below-the-fold thumbnails no longer preload and race the hero.

app/(events)/events/[...filename]/page.tsx
Comment (joshbermanssw) [reply app/(events)/events/[...filename]/page.tsx]: Fixed in 41ba1532. Moved the `if (!filename || filename === "null") notFound()` guard above the `await Promise.all([...])`, so null-slug requests short-circuit before any Tina round-trip.

app/consulting/[filename]/page.tsx
Comment (joshbermanssw) [reply app/consulting/[filename]/page.tsx]: Fixed in 41ba1532. Same change here — the `notFound()` guard now runs before `await Promise.all([newConsultingPageData(filename), consultingPageData(filename)])`.

app/(events)/events/[...filename]/page.tsx
Comment (kulesy) [reply app/(events)/events/[...filename]/page.tsx]: Re-tested locally on the latest commit — the page-body guard is now correct, but `/events/null` still costs 3 Tina round-trips. Source: `generateMetadata` at lines 161–180 runs the same `Promise.all([newEventsPageData, getPreviewEventData, getData])` with no null-check, and Next.js calls it on every request to build the `<head>`.

Dev log on a fresh hit to `/events/null`:
```
Unable to find record content/events/null.mdx       ← still firing
Unable to find record content/eventsv2/null.json    ← still firing
```

Mirror the same guard at the top of `generateMetadata` — `notFound()` isn't legal there, so return `{}` instead:
```ts
const slug = params.filename;
const filename = slug[slug.length - 1];

if (!filename || filename === "null") {
  return {};
}

const [newPage, calendarEvent, oldPage] = await Promise.all([...]);
```

app/consulting/[filename]/page.tsx
Comment (kulesy) [reply app/consulting/[filename]/page.tsx]: Same gap here as the events thread — `generateMetadata` at line 142 calls `newConsultingPageData(filename)` + `consultingPageData(filename)` unguarded, so `/consulting/null` still hits Tina twice via the metadata pass. Same fix shape:
```ts
const filename = params.filename;

if (!filename || filename === "null") {
  return {};
}

const [newPage, oldPage] = await Promise.all([...]);
```

kulesy (approved):
All four review comments resolved and re-verified locally:

- M1a (ImageComponentLayout priority): preload links on /events/ai-workshop-essentials dropped from 8 to 2 (only logo + hero LCP).
- M1b (VideoModal priority): YouTube thumbnail preloads no longer fire on non-LCP video instances.
- Q1a/b (notFound guards): /events/null and /consulting/null both produce zero Tina record lookups in the dev log; generateMetadata is also guarded.

Branch is up to date with main as of merge commit d72e8034. LGTM.

PR #4734, opened 2026-06-05
⚡ Reduce client JS weight on event/content pages

joshbermanssw:
## TL;DR

PageSpeed mobile flags the event pages at **Performance 36** with **1,650 ms Total Blocking Time** and ~3.2 MB of unused JS — driven by heavy client code loaded eagerly on every page, not by page content. This defers or trims the worst offenders (App Insights, framer-motion, GTM) so they stay off the critical path, plus two small content fixes.

**Files to review (7, behaviour-preserving):**

| File | Why |
|---|---|
| `context/app-insight-client.tsx` *(start here)* | Dynamic-import the heavy `applicationinsights-web` core on idle; lightweight React plugin/context stays synchronous so `WebVitals` + error boundary keep working. |
| `components/layout/v2ComponentWrapper.tsx` | Replace framer-motion `useInView` with a small `IntersectionObserver` hook — drops the 2.9 MB package from the wrapper around every v2 block. |
| `app/components/deferred-gtm.tsx` *(new)* | Loads GTM on first interaction or idle, with a guaranteed ~3.5 s fallback. |
| `app/layout.tsx` | Swap `<GoogleTagManager>` for `<DeferredGoogleTagManager>`. |
| `components/blocks/imageComponents/ImageComponentLayout.tsx` | `sizes` cap `min(50vw, 640px)` to stop oversized image fetches. |
| `helpers/getTrimmedEvents.ts` | Guard null/empty slug so no `/events/null` link is prefetched. |
| `app/providers/query-provider.tsx` | `ReactQueryDevtools` → dynamic dev-only import. |

## Why

The page render chain itself is clean — the cost is global/eager client code. Confirmed in-repo: App Insights is instantiated in the root layout on every page; framer-motion is pulled into the wrapper that wraps every block just for one hook; and GTM (which injects Clarity, Hotjar, LinkedIn, FB Pixel) loads `afterInteractive`, contributing ~595 ms of third-party main-thread time.

## How

Each change moves work off the first-load critical path without changing behaviour:

- **App Insights / GTM / devtools** — same code runs, just later (on idle / interaction / dev-only), so they leave the initial bundle and main thread.
- **framer-motion** — the new hook reproduces `useInView(ref, { once: true, margin })` semantics via `IntersectionObserver` `rootMargin`; SSR/no-IO falls back to visible.
- **Images / `/events/null`** — pure correctness: tighter responsive `sizes`, and a slug guard that falls back to `/events`.

## Reviewer notes

- **GTM timing shifts.** Tags fire up to ~3.5 s later (usually sooner). Pageviews still captured for every session via the guaranteed idle/timeout fallback; no code depends on `sendGTMEvent`/`dataLayer`. Dial down `3500` if sub-second tag timing matters.
- **App Insights context intact.** Only the web SDK is lazy; the `ReactPlugin` context is provided immediately, and consumers call it null-safely.
- **Image cap is CSS-px.** Browser still picks the right candidate by DPR, so retina stays crisp; only oversized wide-screen fetches shrink.

## Tests

`tsc --noEmit` clean, ESLint clean on changed files, `jest` 13/13 pass. No runtime/behaviour change to verify beyond the deferral timing above.

## Follow-up

Confirm the named 1.9 MB chunk with `BUNDLE_ANALYSE=true pnpm build` (analyzer already wired in `next.config.mjs`) — these fixes target the most likely contributors (App Insights, framer-motion), but the treemap will show anything else.

### DIFF
--- a/app/components/deferred-gtm.tsx
+++ b/app/components/deferred-gtm.tsx
@@ -1,0 +1,67 @@
+"use client";
+
+import { GoogleTagManager } from "@next/third-parties/google";
+import { useEffect, useState } from "react";
+
+/**
+ * Loads Google Tag Manager off the critical path to reduce Total Blocking Time.
+ *
+ * GTM (and the tags it injects: Clarity, Hotjar, LinkedIn, Facebook Pixel) is
+ * the single largest third-party main-thread cost on first load. None of our
+ * own code calls `sendGTMEvent`/`dataLayer` — every tag fires inside the
+ * container — so it is safe to delay loading until the page is interactive.
+ *
+ * It loads on the first user interaction OR when the browser goes idle,
+ * whichever comes first. The idle callback uses a timeout so the script is
+ * always loaded (within ~3.5s) even for sessions with no interaction, so
+ * pageview tracking is preserved.
+ */
+export function DeferredGoogleTagManager({ gtmId }: { gtmId?: string }) {
+  const [shouldLoad, setShouldLoad] = useState(false);
+
+  useEffect(() => {
+    if (!gtmId || typeof window === "undefined") return;
+
+    let triggered = false;
+    const interactionEvents = [
+      "scroll",
+      "mousemove",
+      "touchstart",
+      "keydown",
+      "click",
+    ];
+
+    const cleanup = () => {
+      interactionEvents.forEach((event) =>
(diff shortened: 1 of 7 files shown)


joshbermanssw:
Closed - outdated and actioned elesewhwre

PR #4741, opened 2026-06-09
update to dynamic imports

joshbermanssw:
fixing unusued JS 
fixing requests to event/null
add priority boolean on imageComponent (media) to set high priorty when needed


isaaclombardssw (approved):
✅

PR #4751, opened 2026-06-10
add null guard to technology card

joshbermanssw:
#4750 

when u instantiate a technology component it has no technology referneces; then you add an item in the list of the cards and it errors out because there is an empty reference

this fixes that


StefStar (approved):
As a Solution architect I approve this PR

PR #4764, opened 2026-06-15
Revert #4761 (event youTubeId) — site went down

isaaclombardssw:
Reverts commit 4ed4bb9af (#4761), which added a `youTubeId` to `content/events-calendar/2026/June-User-Group---Talk-Title-TBA.json`.

The site went down after this commit landed, so reverting to restore it. Re-apply once the cause is understood.


isaaclombardssw:
Closing — the outage was an Azure App Service platform issue (failed to pull mcr.microsoft.com/appsvc/msitokenservice, 401/403), not this content change. Revert is not the fix.

PR #4796, opened 2026-06-24
🐛 Fix JotForm lead-capture 500 in prod

joshbermanssw:
## TL;DR

The lead-capture CTA on `/consulting/react` returns `500 {"message":"JotForm is not configured."}` in prod and PR slots — `/api/lead-capture` reads `process.env.JOTFORM_API_KEY` at runtime, but the key was never plumbed into the running container. This passes the key as a build-arg and, critically, persists it as an `ENV` in the Dockerfile's **runner** stage so the route can read it at runtime.

## Why

`app/api/lead-capture/route.ts` short-circuits to a 500 when `JOTFORM_API_KEY` is unset. The existing build-arg secrets (`TINA_TOKEN`, etc.) are only declared in the **builder** stage, whose ENVs do **not** carry into the final image — so a runtime API route never saw the key. Confirmed against prod:

```
POST https://www.ssw.com.au/api/lead-capture  →  500
{"message":"JotForm is not configured."}
```

## How

1. `template-build.yml` passes `JOTFORM_API_KEY=${{ secrets.JOTFORM_API_KEY }}` as a build-arg (both prod and PR-slot builds reuse this workflow).
2. `Dockerfile` builder stage declares `ARG JOTFORM_API_KEY` (+ `ENV`) for parity with the other secrets.
3. `Dockerfile` **runner** stage re-declares `ARG`/`ENV JOTFORM_API_KEY` — this is the fix: it's what makes the value survive into the final image the App Service runs.

## Reviewer notes

- **Runner ENV is the fix.** Builder-stage ENVs don't reach the final image; the runner-stage `ENV` is what exposes it at runtime.
- **`secrets: inherit` covers it.** Both callers inherit secrets, so the new build-arg resolves without declaring it in `template-build.yml`'s `secrets:` block.
- **Tradeoff: baked into image.** This stores the key in the ACR image layers (chosen over the Key Vault `appSettings` pattern). Acceptable per the same approach used for other build-arg secrets.

## Prerequisite

`JOTFORM_API_KEY` must exist as a GitHub Actions secret (added). No Key Vault / infra change needed for this approach.


wicksipedia (approved):

18-th (approved):
LGTM

PR #4798, opened 2026-06-24
🐛 Send consulting leads to CRM via Power Automate

joshbermanssw:
## TL;DR

The React consulting lead form wrote to JotForm via the API, but API submissions don't fire JotForm's webhook — so the JotForm→Power Automate→CRM flow never ran and leads never reached Dynamics. `/api/lead-capture` now writes the lead to JotForm (record) **and** POSTs it to a dedicated Power Automate HTTP-trigger flow that creates the CRM lead + sends the notification email.

## Why

- JotForm webhooks only fire on real form submissions, not API writes, and the form's captcha blocks a server-side form post — so the existing connector-triggered flow was unreachable from our form.
- A parallel HTTP-trigger flow (deployed via the SSW.CRM pipeline) gives us a trigger we control.

## How

1. **`route.ts`** — keeps the JotForm API write for the record, then POSTs the lead JSON to `POWER_AUTOMATE_LEAD_WEBHOOK_URL`; failures there are logged but don't fail the user's submission.
2. **`leadCapture.tsx`** — sends a clean semantic payload (`name`, `email`, `company`, `phone`, `location`, `hearAboutUs`, `message`, `landingPageUrl`); the route maps to JotForm qids.
3. **`.env.example`** — documents the new `POWER_AUTOMATE_LEAD_WEBHOOK_URL`.

## Notes

- Requires `POWER_AUTOMATE_LEAD_WEBHOOK_URL` set per environment (the HTTP-trigger flow's URL — UAT/PROD differ).
- Pairs with the SSW.CRM solution work that deployed the HTTP-trigger flow to UAT + PROD.


isaaclombardssw (approved):
✅ LGTM

PR #4777, opened 2026-06-18
Consulting Page for React (A|B Testing) 

joshbermanssw:
A/B rebuild of the React consulting page as a TinaCMS v2 page, built from a new set of `v3` blocks. The route only resolves to the new design once `consultingv2/react.json` exists, so the old page stays live until cutover. Related: #4775.

## New `v3` blocks

- **Hero** — split hero: brow, title, description, button row, media slot (image or built-in animated React composition)
- **Logo Carousel** — client logo marquee
- **Feature Steps** — numbered (01/02/03) feature/problem columns
- **Process** — numbered process timeline
- **Statistics** — headline stat figures grid
- **Testimonials** — carousel with line-by-line clip text reveal on switch
- **Stack Cards** — up to 4 tech-stack cards with arrow links
- **FAQ** — accordion (grid-rows height animation)
- **CTA** — closing call-to-action
- **Lead Capture** — multi-step quiz feeding a JotForm

Shared: new background options on every v3 block — dotted **grid overlay** and **red glow** — plus hex codes in the Tina colour picker.

## Links

- [Figma — SSW Website Redesign (Alex Blum)](https://www.figma.com/design/cXnKgs11UOW05fxJ1lwm35/SSW-Website-Redesign---Alex-Blum?node-id=10883-3253&t=4TKaUImPaqXwMy3T-0)

PR #4805, opened 2026-06-26
Protect React consulting lead-capture endpoint (Turnstile) + review fixes

### LINKED ISSUE #4823: React consulting lead-capture endpoint is unprotected
The `/api/lead-capture` route added with the React consulting v3 page (#4777) is a public endpoint that spends the JotForm API key and fires the Power Automate CRM flow on every call, with nothing gating it:

- Middleware only sets headers — no auth, origin check, or rate limit
- The `*.azurewebsites.net` origin and preview slots bypass Cloudflare entirely
- JotForm's own limit is just a daily API-call quota (hitting it blocks legit submissions too), and the API path bypasses the form's built-in CAPTCHA

So it can be scripted to spam leads into both JotForm and the CRM.

**Fix:** require a Cloudflare Turnstile token (we already have a Cloudflare account; Turnstile is free) before the route does any JotForm/CRM work — fail closed.

While here, also restore a few fixes prepared during review of #4777 that didn't land: SEO metadata on the React page, process step zero-padding, Tailwind token usage, and the removed next/image `objectFit` prop.

Affected routes: `/consulting/react`

isaaclombardssw:
Fixed #4823

Follow-up to #4777 (React consulting v3 page), which merged without the protection and review fixes that were prepared during review but never pushed. This adds them on top of current main.

## What & why

**Lead-capture protection (main change)**
`/api/lead-capture` is a public endpoint that spends the JotForm API key and fires the Power Automate CRM flow on every call, with nothing gating it (middleware only sets headers; the `*.azurewebsites.net` origin/preview slots bypass Cloudflare). So it can be scripted to spam leads into both JotForm and the CRM. It now requires a valid **Cloudflare Turnstile** token before doing either.

- `@marsidev/react-turnstile` widget (interaction-only) on the contact step of the shared `LeadCaptureForm`, so both the React page and the lead-capture modal are protected
- `validateTurnstile` server helper → Turnstile siteverify; logs a missing secret or a non-OK siteverify response (still fails closed) so an unprovisioned server isn't silently rejecting real users
- Fail-closed: no/invalid token → 400/403 before any JotForm or CRM call

**Review fixes (still missing from main)**
- Restore SEO (title/description/canonical/OG) on the React page — lost when `react.mdx` was replaced by `react.json`
- Zero-pad process step numbers (`01, 02…`, was `" 1"`)
- Use `sswBorder` / `border-0.75` tokens instead of arbitrary `border-[#212121]` / `border-t-[0.75px]`
- Replace removed next/image `objectFit` prop with `object-contain` class

## Requires before it works
Add two **GitHub Actions repository secrets** (Settings → Secrets and variables → Actions), matching how `JOTFORM_API_KEY` is handled:
- `NEXT_PUBLIC_TURNSTILE_SITE_KEY` (Turnstile site key)
- `TURNSTILE_SECRET_KEY` (Turnstile secret key)

The build wiring (reusable workflow + Dockerfile) is included here and the secrets are `required: false`, so builds don't break before the values are set — the form fails closed (won't submit) until both exist. No Key Vault / bicep change needed; PR preview slots get the keys via their own build.

Affected routes: `/consulting/react`


18-th (approved):
LGTM

PR #4806, opened 2026-06-26
✨ Rebuild the homepage with new V3 content blocks

joshbermanssw:
## TL;DR

Rebuilds the SSW homepage on a new set of `<V3>` TinaCMS blocks — a full-bleed hero carousel, services / case-study / people sections, an events block with a live countdown, and an interactive WebGL globe of our offices. Content is driven by `content/pagesv2/home.json` and rendered through the v3 blocks registry at `/`.

## New components

- [ ] **`<V3> Hero Box`** — full-bleed rounded hero carousel with an SSW-red duotone tint, legibility scrim, and scroll-down scoop
- [ ] **`<V3> Image Cards`** — services grid (desktop) / finite carousel with a "+ more" end cap (mobile)
- [ ] **`<V3> Video Hero`** — full-width muted video-background hero
- [ ] **`<V3> Featured Products`** — case-studies highlight plus a supporting list
- [ ] **`<V3> People Carousel`** — team-member carousel
- [ ] **`<V3> Video Feature`** — "why choose" video with highlights and recognition badges
- [ ] **`<V3> Events`** — upcoming events pulled from the calendar with a live countdown
- [ ] **`<V3> Globe`** — offices accordion beside an interactive `cobe` WebGL globe
- [ ] **Lead Capture** — shared lead-capture form + modal wired to the block CTAs

Supporting additions: shared `SectionHeader` / `CarouselMoreCard` helpers and a `useMediaQuery` hook.

- Affected routes: `/` (homepage)


components/blocks/v3/globe/globe.tsx
```diff
@@ -0,0 +1,164 @@
...
+                      }
+                    />
+                  ))}
+                </div>
+              )}
+            </div>
+
+            {/* Right: 3D globe stage. */}
+            {offices.length > 0 && (
+              <div className="relative z-0 hidden lg:col-span-7 lg:block">
```
joshbermanssw, line 157: **[P1] Avoid mounting the hidden WebGL globe on mobile**

`hidden` only changes CSS; `OfficeMap` still mounts on every mobile device, creates a 16,000-sample WebGL globe, and starts a perpetual `requestAnimationFrame` loop even though the result can never be seen. Please gate the component with the desktop media query (and ideally pause it while off-screen) so mobile visitors do not pay the CPU/GPU and battery cost for this below-fold decoration.

content/pagesv2/home.json
```diff
@@ -0,0 +1,515 @@
...
+          "altText": "AI-HackDay-SYD",
+          "imageSource": "/images/background/banner-custom-trainings.jpg",
+          "imageHeight": 876,
+          "imageWidth": 3684
+        },
+        "registerText": "Register Now",
+        "buttons": [
+          {
+            "buttonText": "Register Now",
+            "buttonLink": "/events/2026/ai-hack-day---sydney",
```
joshbermanssw, line 412: **[P1] Do not pin the dynamic featured-event CTA to one event**

The block dynamically renders the first future calendar event, but `getFeaturedButton` gives this configured URL precedence over that event's generated URL. As soon as AI Hack Day is no longer the first upcoming event, the title/date will describe one event while Register navigates to another; the image override will also remain AI-HackDay-specific. Leave these overrides blank or derive all featured display data from the selected event.

components/blocks/v3/events/events.tsx
```diff
@@ -0,0 +1,476 @@
...
+}
+
+function getEventLocation(event) {
+  if (!event?.city) return "";
+  return event.city === "Other" ? event.cityOther : event.city;
+}
+
+function getEventTime(event) {
+  if (!event?.startDateTime) return "";
+  return dayjs(event.startDateTime).format("h:mma AEST");
```
joshbermanssw: **[P1] Format the event in the advertised time zone**

This formats in the runtime's local zone and merely appends the literal `AEST`. On Vercel that can show UTC as AEST, and Sydney is AEDT during daylight saving. Please use the timezone plugin/`Intl.DateTimeFormat` with `timeZone: "Australia/Sydney"` and emit the actual zone name.

components/blocks/videoBackground.tsx
```diff
@@ -11,13 +12,30 @@ const VideoBackground: React.FC<VideoBackgroundProps> = ({
...
+
   return (
     <video
+      ref={videoRef}
       className="absolute h-full min-w-full object-cover transition-opacity duration-1000 z-bgVideo"
       playsInline
       autoPlay
       muted
       loop
+      preload="auto"
```
joshbermanssw: **[P2] Do not eagerly preload this below-fold video**

The homepage places this section after three other blocks and the supplied MP4 is 4.8 MB. `preload="auto"` asks browsers to fetch it during the critical initial load, competing with the hero/LCP resources. Please use `metadata`/`none` and start loading when the section approaches the viewport (or make preload caller-controlled for genuinely above-fold uses).

components/blocks/v3/imageCards/imageCards.tsx
```diff
@@ -0,0 +1,192 @@
...
+import Image from "next/image";
+import Link from "next/link";
+import { BsArrowUpRight } from "react-icons/bs";
+import { tinaField } from "tinacms/dist/react";
+
+function ImageCard({ card, cardBackgroundClass, showBorder }) {
+  const inner = (
+    <div
+      className={cn(
+        "group relative flex h-full flex-col overflow-hidden rounded-card border-0.75 border-sswBorder transition",
```
joshbermanssw: **[P2] Make the border toggle effective**

The unconditional base classes already include `border-0.75 border-sswBorder`, so adding the same classes when `showBorder` is true changes nothing. The new homepage explicitly sets `showBorder: false` but will still render borders. Move the border classes into this conditional (and use `border-0` for the false branch if needed).

components/blocks/v3/videoHero/videoHero.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+      {/* Darkening overlay for legibility */}
+      <div
+        aria-hidden="true"
+        className="absolute inset-0 bg-black/60 z-videoMask"
+      />
+
+      <div className="relative flex w-full max-w-4xl flex-col items-center px-4 text-center z-content">
+        {data?.heading && (
+          <h1
```
joshbermanssw: **[P2] Preserve a single page-level heading**

The first Hero Box already renders the homepage `h1`, so this later section adds a second `h1` midway through the document. Render this block's heading as `h2` (or make the level explicit from context) to preserve the page outline for screen-reader and SEO consumers.

package.json
```diff
@@ -113,6 +114,7 @@
     "sharp": "^0.34.3",
     "ssw.megamenu": "4.13.8",
     "stale-while-revalidate-cache": "^3.4.0",
+    "svg-dotted-map": "^2.1.0",
```
joshbermanssw: **[P3] Remove the abandoned dotted-map dependency**

`svg-dotted-map` is only imported by the newly added `components/ui/dotted-map.tsx`, and that component has no consumers; the shipped globe uses `cobe` instead. Please remove this dependency, its lockfile entry, the unused 196-line component, and the now-unused Magic UI registry configuration so the PR does not retain an abandoned implementation.

components/layout/footer/footer.tsx
```diff
@@ -86,7 +86,9 @@ export const Footer = () => {
           ))}
         </div>
 
-        <div className="mt-4 flex flex-col gap-4 text-sm md:flex-row md:items-center md:justify-between">
+        <hr className="hidden sm:block my-2 h-px border-gray-700 md:my-8" />
```
joshbermanssw: **[P1] Keep the lint check green**

These two added class strings fail both Prettier and `tailwindcss/classnames-order`, which is why the PR's `lint-code` check is red. Running the formatter should reorder this line to `my-2 hidden h-px border-gray-700 sm:block md:my-8` and the following container to `mt-4 flex flex-col gap-4 text-sm sm:mt-0 ...`.

joshbermanssw (commented):
Reviewed the complete PR, including the homepage data migration, all new V3 blocks, shared component changes, dependencies, and local lint output. I found the inline issues below; the current build and tests pass, while `pnpm lint` fails on the footer additions.

content/pagesv2/home.json
joshbermanssw: < IGNORE >

components/blocks/v3/globe/globe.tsx
```diff
@@ -0,0 +1,169 @@
...
+      >
+        <span className="text-lg text-white">{office?.name}</span>
+        <FiChevronDown
+          className={cn(
+            "size-5 shrink-0 text-white transition-transform duration-300",
+            isOpen && "rotate-180"
+          )}
+        />
+      </button>
+
```
joshbermanssw, line 41: **[P2] Remove collapsed office links from keyboard navigation**

The closed state only animates the grid row to zero and opacity to zero; its email, phone, and site anchors remain focusable. Keyboard users will tab through invisible controls for every collapsed office. Please conditionally render the panel, apply `hidden` after the close transition, or use `inert`/an accordion primitive so closed content is removed from the focus order.

components/blocks/v3/events/events.tsx
```diff
@@ -0,0 +1,498 @@
...
+                >
+                  <EventListItem event={event} />
+                </div>
+              ))}
+            </div>
+
+            {seeMoreButtons.length > 0 && (
+              <div className="flex justify-end px-4 lg:px-0">
+                <ButtonRow
+                  className="mt-0 hidden justify-end lg:block"
```
joshbermanssw: **[P2] Keep the Events CTA reachable on mobile**

Unlike the other sections, Events has no mobile `CarouselMoreCard`, while this only `See More Events` control is explicitly hidden below `lg`. Mobile visitors therefore see the limited event list with no path to the full events page. This CTA is also nested under `eventCards.length > 0`, so it disappears at every breakpoint when only the featured event is returned. Render it independently of the card count and provide a mobile-visible action.

components/blocks/v3/heroBox/heroBox.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  function slideRight() {
+    setActiveSlide((current + 1) % slides.length);
+  }
+
+  const prevAndNextSlideButtons = slides.length > 1 && (
+    <div className="absolute bottom-1/2 left-1/2 flex w-full -translate-x-1/2 items-center justify-center gap-2">
+      <button
+        type="button"
+        aria-label="Scroll to next section"
```
joshbermanssw: **[P2] Give the carousel controls accurate accessible names**

Both arrow buttons are announced as “Scroll to next section”, even though one selects the previous hero slide and neither scrolls the page. Please label them “Previous slide” and “Next slide” so screen-reader users can distinguish the actions.

joshbermanssw (commented):
Follow-up review of the remaining PR surface after accepting the pinned featured-event behavior. I found three additional accessibility/responsive issues inline. No approval submitted.

components/blocks/v3/globe/globe.tsx
joshbermanssw: Gated the `OfficeMap` mount behind a `useMediaQuery("(min-width: 1024px)")` so the WebGL globe only mounts on the desktop viewports that show it, and added an `IntersectionObserver` that pauses its `requestAnimationFrame` loop while it is scrolled off-screen. Fixed in f22d9fe4.

components/blocks/v3/events/events.tsx
joshbermanssw: `getEventTime` now formats the start time in `Australia/Sydney` via dayjs `.tz()` and emits the real zone label (AEST/AEDT) from `Intl.DateTimeFormat`, rather than hardcoding the `AEST` suffix. Fixed in f22d9fe4.

components/blocks/videoBackground.tsx
joshbermanssw: `preload` now defaults to `none` and is a caller-controllable prop; loading and playback are deferred until the section nears the viewport via an `IntersectionObserver`, so the below-fold MP4 no longer competes with the hero/LCP resources. Fixed in f22d9fe4.

components/blocks/v3/imageCards/imageCards.tsx
joshbermanssw: Moved the border classes out of the base and into the `showBorder` conditional (with `border-0` on the false branch), so the toggle now actually adds/removes the border. Fixed in f22d9fe4.

components/blocks/v3/videoHero/videoHero.tsx
joshbermanssw: Rendered this section heading as an `h2` so the page keeps a single `h1` (the Hero Box). Fixed in f22d9fe4.

package.json
joshbermanssw: Removed the `svg-dotted-map` dependency, its `pnpm-lock.yaml` entries, the unused `components/ui/dotted-map.tsx`, and the now-unused Magic UI registry config from `components.json`. Fixed in f22d9fe4.

components/layout/footer/footer.tsx
joshbermanssw: Reordered the Tailwind classes to satisfy Prettier and `tailwindcss/classnames-order`; the touched files pass `eslint` and `prettier --check`. Fixed in f22d9fe4.

components/blocks/v3/globe/globe.tsx
joshbermanssw: Closed accordion panels now get `inert={!isOpen}` (React 19), which removes their email/phone/site links from the tab order and the accessibility tree while keeping the grid collapse animation. Fixed in 1a62797c.

components/blocks/v3/events/events.tsx
joshbermanssw: Moved the See More CTA out of the `eventCards.length > 0` block so it renders independently of the card count, and made it visible at every breakpoint (centered on mobile, right-aligned on desktop) so featured-only/mobile renders keep a path to `/events`. Fixed in 1a62797c.

components/blocks/v3/heroBox/heroBox.tsx
joshbermanssw: Relabeled the arrows `Previous slide` and `Next slide` instead of both announcing `Scroll to next section`. Fixed in 1a62797c.

isaaclombardssw:
The hero section notch used to have a down arrow that fit the space nicely, it looks strange to have the notch without anything living there.

My recommendation would be bring back the down arrow for non carousel-case

Screenshot 2026-07-10 at 12 21 03 pm

**Figure: original brief**

Screenshot 2026-07-10 at 12 20 40 pm

**Figure: notch in staging**

content/consultingv2/react.json
```diff
@@ -178,7 +178,7 @@
           "description": "Across all industries"
         },
         {
-          "figure": 30,
+          "figure": 27,
           "figureSuffix": "+",
           "heading": "Years experience",
           "description": "Trusted since 1999"
```
isaaclombardssw, line 184: https://www.ssw.com.au/company/history

1990 🤯

public/images/videos/ai_solutions_background.mp4
isaaclombardssw, line 1: rename this

public/images/new-homepage-background-assets/fpe.png
isaaclombardssw, line 1: this and other homepage background assets should probably be optimised, 3.39 MB is big. They should each be maybe 400kb or less WebP or JPG

Next/image helps but moreso repo hygeine (and cold-cache means we re-encode the smaller version every time).

Probably a bit nitpicky

content/pagesv2/home.json
```diff
@@ -0,0 +1,515 @@
...
```
isaaclombardssw, line 412: How are we going to know when this gets stale?

components/blocks/v3/heroBox/heroBox.tsx
```diff
@@ -0,0 +1,187 @@
...
+          />
+        ))}
+      </div>
+    </div>
+  );
+
+  const scoopColor = backgroundOptions.find(
+    (option) => option.reference === data?.background?.backgroundColour
+  )?.hex;
+
```
isaaclombardssw: Does this need a fallback?

hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,24 @@
...
+      setMatches(event.matches);
+
+    setMatches(mediaQuery.matches);
+    mediaQuery.addEventListener("change", handleChange);
+
+    return () => mediaQuery.removeEventListener("change", handleChange);
+  }, [query]);
+
+  return matches;
+}
```
isaaclombardssw: possibly the same as pre-existing  usehooks-ts but double check

components/blocks/v3/shared/carouselMoreCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+      target={newTab ? "_blank" : undefined}
+      rel={newTab ? "noopener noreferrer" : undefined}
+      className={cn(
+        "group flex h-full flex-col items-center justify-center gap-4 rounded-card p-6 text-center !no-underline",
+        className
+      )}
+    >
+      <span className="flex size-12 shrink-0 scale-100 items-center justify-center rounded-full bg-white text-black transition-all duration-300 ease-in-out group-hover:rotate-45 group-hover:scale-125">
+        <BsArrowUpRight className="size-5" />
+      </span>
```
isaaclombardssw: should this be it's own atom and shared with other components with the same interaction

isaaclombardssw (changes requested):
Pretty clean just a couple nits

content/consultingv2/react.json
joshbermanssw: technically the company was listed in 1995 (looked at this w/ adam and uly) but for this sakes ill put 30+

content/pagesv2/home.json
joshbermanssw: This `featuredEvent` is a manual pin — its image and Register button intentionally override the live calendar event, so it'll need swapping by hand once AI Hack Day passes. No automated staleness signal for now; we're keeping it manually maintained.

components/blocks/v3/statistics/statistics.tsx
```diff
@@ -60,7 +60,7 @@ export default function V3Statistics({ data }) {
                 {stat?.figureSuffix && (
                   <span
                     data-tina-field={tinaField(stat, "figureSuffix")}
-                    className="ml-1 text-3xl font-light text-white lg:text-3xl xl:text-4xl"
+                    className="ml-1 text-4xl font-light text-white lg:text-3xl xl:text-4xl"
```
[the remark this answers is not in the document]
joshbermanssw: Intentional — the figure it sits beside dips the same way (`text-6xl lg:text-5xl xl:text-6xl`), so the `+` suffix shrinks at `lg` to stay proportional and returns to full size at `xl`. A constant suffix size would look oversized against the shrunken number at `lg`.

content/consultingv2/react.json
isaaclombardssw: lol

isaaclombardssw (approved):
✅

PR #4809, opened 2026-06-29
🧹 Remove /healthcheck route and footer link

joshbermanssw:
cc: @tiagov8 @adamcogan 

As per email from Tiago

```
See {{ RULE LINK }}

Please see if you can follow the HealthCheck rule (written by Toby)
If not, remove the link and page from the SSW Website
```


0xharkirat (approved):
LGTM

PR #4819, opened 2026-06-29
update preview url 

joshbermanssw:
Updates preview URL as per Adams request so it takes u to the site just with no admin editor (same route)


0xharkirat (approved):
Regex is checked by Claude Code - LGTM

PR #4825, opened 2026-07-01
fix(v2ComponentWrapper): don't set background-image when URL is undefined

### LINKED ISSUE #2870: ✨ New Template - Packaged Services
from email subject `Re: 👑 IMPORANT -  .NET MAUI Mobile Application: 3-day Assessment – Microsoft Azure Marketplace`

This is where all the new Azure Marketplace consulting services can live.

```[tasklist]
### Tasks
- [ ] Create the new collection schema
- [ ] Create the new page template
- [ ] MegaMenu - Add it somewhere under Services. could be a button next to View All Services, or a 3rd column
- [ ] Get a checked by from @samsmithssw @miichaelsmedley @uly1
```

0xharkirat:
Closes #2870

## Summary

The v2 block wrapper unconditionally sets an inline `background-image: url(...)` even when no image is configured, producing:

\`\`\`html
<section style=\"background-image:url(undefined);background-size:cover;...\">
\`\`\`

The browser resolves \"undefined\" as a relative URL and fires requests like \`/consulting/<slug>/undefined\`, which is what was causing the mystery \`GET /consulting/undefined\` and \`GET /consulting/null\` hits in dev logs (and the \"Unable to find record content/consultingv2/undefined.json\" GraphQL warnings) that we tracked down in [#4714](https://github.com/SSWConsulting/SSW.Website/pull/4714).

## Change

One-line guard in \`components/layout/v2ComponentWrapper.tsx\` — skip the \`backgroundImage\` style when \`data.background?.backgroundImage\` is falsy. Colour classes and bleed image still work.

## Test plan

- [ ] Any block without a background image no longer emits \`background-image: url(undefined)\` in the DOM
- [ ] Blocks with a background image still render it correctly
- [ ] No \`/consulting/undefined\` requests in the network tab or dev server logs

### DIFF
--- a/components/layout/v2ComponentWrapper.tsx
+++ b/components/layout/v2ComponentWrapper.tsx
@@ -114,10 +114,10 @@ const V2ComponentWrapper = ({
           !isInInitialViewport && isInView && "opacity-100"
         )}
         style={
-          data.background?.bleed
+          data.background?.bleed || !data.background?.backgroundImage
             ? {}
             : {
-                backgroundImage: `url(${data.background?.backgroundImage})`,
+                backgroundImage: `url(${data.background.backgroundImage})`,
                 backgroundSize: "cover",
                 backgroundRepeat: "no-repeat",
                 backgroundPosition: "center",


0xharkirat:
NOt required as #4826 fixed the issue

PR #4782, opened 2026-06-18
V2 card spacing + feature-list bullet-icon polish

isaaclombardssw:
Closes #4779

From Tiago's design feedback on the AI for Business Leaders event page (PR #4760). Three small shared-component tweaks:

- **#3 card padding** — Glass card style (`cardOptions.tsx`) gains `pb-6 lg:pb-8` so cards have more space at the bottom.
- **#6 feature-list spacing** — imageTextBlock feature grid gains `my-4 gap-3` for vertical breathing room against the heading/description.
- **#7 bullet icon colour** — feature-list (`ListItem`) bullet icons go from white to `sswRed`. (Card icons were already red.)

**Draft — site-wide, for Tiago to confirm:** all three touch shared V2 components used across consultingv2 / eventsv2 / etc. Open calls: exact card padding value; whether bullet icons should be red everywhere or only on dark sections. Spot-check other V2 pages before merge.

Preflight reviewed twice; lint clean; verified rendering on /consulting/azure.

PR #4781, opened 2026-06-18
Eyebrow (topLabel) restyle — uppercase + SSW red

isaaclombardssw:
Closes #4778

Restyles the eyebrow / topLabel (`components/blocksSubtemplates/iconLabel.tsx`) to uppercase + wide tracking in `sswRed`, with the optional icon tinted to match. From Tiago's design feedback on the AI for Business Leaders event page (PR #4760, item #4).

**Draft — open decision for Tiago:** `IconLabel` renders *every* eyebrow/topLabel across all imageTextBlock and cardCarousel sections site-wide, so this restyles them all. This PR implements the global version (simplest, and likely the design-system direction). If we'd rather scope it to specific sections, say so and I'll add a variant/prop instead.

Also worth a glance: with `dark:text-gray-300` removed the label is red in both light and dark — confirm contrast on dark eyebrow backgrounds.

Preflight reviewed twice; lint clean; verified rendering on /consulting/web-applications.

PR #503, opened 2023-04-12
Update Infra - adding key vault to app config

amankumarrr:
closes #502 

- [ ] We need to make sure all PRs role assignments of ACR must be removed before running this

PR #4843, opened 2026-07-08
fix(infra): make ACR pull role assignment idempotent per principal

### LINKED ISSUE #4844: PR 'Deploy to slot' fails: ACR role assignment not idempotent (RoleAssignmentUpdateNotPermitted)
### Symptom
The **Deploy to slot** step fails on PR environments (and has surfaced on `main`) with:

\`\`\`
RoleAssignmentUpdateNotPermitted: Tenant ID, application ID, principal ID, and scope are not allowed to be updated.
target: .../deployments/pr-<n>-acrRoleAssignment-...
\`\`\`

### Root cause
`infra/acrRoleAssignment.bicep` names the ACR "AcrPull" role assignment `guid(acr.id, roleIdMapping[roleName], slotName)` - keyed on the slot name, **not** the principal. When a PR staging slot is torn down and recreated (slot cleanup), its system-assigned identity comes back with a new `principalId`, but the role-assignment name is unchanged, so the deploy submits the same-named assignment with a different principal, which Azure rejects as an illegal in-place update.

Confirmed against Microsoft's "Troubleshoot Azure RBAC" doc (updated 2026-05-24), which documents this exact scenario and prescribes keying the name on `principalId`. `infra/keyVaultRoleAssignment.bicep` already does this; the ACR one was the outlier. Latent since #503 (2023-05-08).

### Fix
Key the assignment name on `principalId` (`guid(acr.id, principalId, roleIdMapping[roleName])`) so re-deploys are idempotent per principal. PR: #4843.

### Note
Contributing factor: the App Service plan's staging-slot cap (max ~10) fills up with open-PR environments, which is what forces slot recycling in the first place. Cleaning up stale PR slots is a related follow-up.

isaaclombardssw:
Closes #4844

## What

`infra/acrRoleAssignment.bicep` named the "ACR Pull" role assignment with `guid(acr.id, roleIdMapping[roleName], slotName)` — keyed on the **slot name**, not the **principal**. Change it to `guid(acr.id, principalId, roleIdMapping[roleName])` (mirroring `keyVaultRoleAssignment.bicep`, which already keys on `principalId`). `slotName` becomes unused, so the param and its pass-through in `create-pr-environment.bicep` are removed.

## Why

When a PR slot is torn down and recreated (slot cleanup), the new slot's managed identity comes back with a **new `principalId`**, but the role-assignment name is unchanged. The deploy then tries to rewrite the same-named assignment with a different principal, which Azure refuses:

```
RoleAssignmentUpdateNotPermitted: Tenant ID, application ID, principal ID, and scope are not allowed to be updated.
```

That's what fails the **Deploy to slot** step on PRs whose environment has been recycled (e.g. #4781, #4782, and the TinaCMS content PRs) and on `main`. Keying the name on `principalId` makes it idempotent: same principal -> same GUID (re-runs are no-ops), new principal -> new GUID (a fresh assignment instead of an illegal update). The orphaned old assignment is harmless (its principal no longer exists).

## History

Latent since `infra/acrRoleAssignment.bicep` was added in #503 (2023-05-08). It only bites once a slot with the same name is recreated with a new identity, which is why older/recycled PR environments hit it now rather than on first deploy.

## Verification

- `az bicep build infra/create-pr-environment.bicep` compiles clean.
- `keyVaultRoleAssignment.bicep` already uses the principal-keyed pattern; no change needed there.

### DIFF
--- a/infra/acrRoleAssignment.bicep
+++ b/infra/acrRoleAssignment.bicep
@@ -1,7 +1,6 @@
 param roleName string
 param principalId string
 param acrName string
-param slotName string
 
 var roleIdMapping = {
   'ACR Pull': '7f951dda-4ed3-4680-a7ca-43fe172d538d'
@@ -13,7 +12,7 @@ resource acr 'Microsoft.ContainerRegistry/registries@2021-09-01' existing =  {
 
 
 resource addRoleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
-  name: guid(acr.id, roleIdMapping[roleName], slotName)
+  name: guid(acr.id, principalId, roleIdMapping[roleName])
   scope: acr
   properties: {
     roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', roleIdMapping[roleName])
--- a/infra/create-pr-environment.bicep
+++ b/infra/create-pr-environment.bicep
@@ -47,7 +47,6 @@ module acrRoleAssignment 'acrRoleAssignment.bicep' = {
     acrName: acrName
     principalId: appServiceSlot.outputs.slotPrincipalId
     roleName: 'ACR Pull'
-    slotName: slotName
   }
 }
 


isaaclombardssw:
## 🧠 Sanity Checker: Confirmed (high confidence)

Verified the root cause this PR fixes: the ACR "AcrPull" role assignment was named `guid(acr.id, role, slotName)` (no principalId), so when a PR slot is recreated with a new system-assigned identity the same-named assignment is submitted with a new principalId, which Azure rejects as `RoleAssignmentUpdateNotPermitted`. Two independent agents ran, one confirming via authoritative sources, one trying to refute.

**Confirm agent (Confirmed):** Microsoft's "Troubleshoot Azure RBAC" page documents this exact scenario (delete + recreate a managed identity gives a new principal ID; redeploying the same-named assignment fails) and prescribes this exact fix (`guid(scope, principalId, roleDefinitionId)`). All logical steps backed by live docs.

**Refute agent (could not break it, Confirmed):** Two caveats, neither overturns the fix. The error can also be triggered by changing scope or roleDefinitionId, but both are fixed here so principalId is the only variable. And keying on principalId leaves the old assignment as a harmless orphan that is worth periodic cleanup.

### Sources
- Troubleshoot Azure RBAC (updated 2026-05-24): https://learn.microsoft.com/en-us/azure/role-based-access-control/troubleshooting
- Bicep RBAC scenarios (updated 2025-12-23): https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/scenarios-rbac
- App Service managed identity: https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity

joshbermanssw (approved):

PR #4845, opened 2026-07-09
Update pr-deploy-command

joshbermanssw:
Update the pr-deploy-command to have failues commented earlier and remove the failing lighthouse runs on the staging deployments

PR #4846, opened 2026-07-09
🐛 Fix /deploy workflow rejected at startup after #4845

joshbermanssw:
## TL;DR

Since #4845 merged, GitHub couldn't parse `pr-deploy-command.yml`, so it showed as a raw file path in the Actions tab and every event produced a 0-second startup-failure — commenting `/deploy` silently did nothing. This restores it to a valid, parseable workflow (`actionlint` clean).

## Why

Two independent workflow-level errors from #4845 each made GitHub reject the whole file at load time:

- An empty `${{ }}` sat inside a comment in the `notify-failure` run block. Actions templates the run script *before* pwsh sees it, ignoring the `#`, and an empty expression is a syntax error.
- `prLighthouseInsights` read `needs.validate.*` in its `with:` block but only declared `needs: [pr-deploy]` — referencing an undeclared dependency.

Also added the required `description` to the `deploy` composite action (`action.yml` needs both `name` and `description`), which #4845 touched.

Once merged, the sidebar entry returns to its proper name, **"PR - deploy to slot on /deploy"**.

## Links

- [Actions runs for this workflow](https://github.com/SSWConsulting/SSW.Website/actions/workflows/pr-deploy-command.yml)
- #4845


18-th (approved):
LGTM

PR #4867, opened 2026-07-14
💄 Add light/dark mode toggle to the v3 homepage

joshbermanssw:
## TL;DR

The v3 homepage was dark-only. This adds a light/dark toggle **scoped to the homepage** and migrates the homepage blocks off hardcoded dark colours (`text-white`, `text-gray-*`, `bg-sswCard`, `border-sswBorder`) to theme tokens (`text-foreground`, `text-muted-foreground`, `border-hairline`) so light mode renders correctly — dark mode is byte-for-byte unchanged.

## How

- `HomeThemeProvider` + `HomeThemeShell` apply `.dark` to a homepage-scoped wrapper (not `<html>`), so Tailwind `dark:` variants and the design-token CSS vars resolve from that ancestor.
- The nav's sun/moon toggle flips/persists the choice (`localStorage`); a first-time visitor with no saved choice follows their OS `prefers-color-scheme` (system default) — dark is only the SSR seed, corrected before first paint by an inline script so there's no flash.
- Each migrated block keeps its exact dark appearance via `dark:` variants; only the light path is new.

## Reviewer notes

- **Scoped, not global.** `.dark` lives on a homepage wrapper only — other routes (including the always-`.dark` consulting v2 pages that share these v3 blocks) are unaffected.
- **Media stays dark by design.** videoHero, heroBox, image-overlay cards, the countdown, SSW-red badges and the WebGL globe read as dark panels in both themes — their text sits over video/photos.
- **Toggle placement is temporary.** Bottom-left floating button until the megamenu header extension point is sorted.

## Status

Draft — functionally complete and lint-clean; opening for a preview deploy + visual review.


components/blocksSubtemplates/tinaFormElements/colourOptions/blockBackgroundOptions.tsx
```diff
@@ -1,61 +1,101 @@
...
   {
     name: "Dark Gray",
-    classes: "bg-gray-950 text-white",
+    classes: "bg-gray-50 text-foreground dark:bg-gray-950 dark:text-white",
     reference: 4,
-    editorClasses: "bg-[#222222] text-white",
     hex: "#030712",
+    lightHex: "#f9f9f9",
+    lightPreviewClasses: "bg-[#f9f9f9] text-black",
+    darkPreviewClasses: "bg-[#030712] text-white",
```
isaaclombardssw: This dark preview reads `#030712`, but this repo overrides Tailwind's grays, so `bg-gray-950` actually renders **#222222** on the page (`Gray` / `bg-gray-900` = **#333333**, and the `from-gray-900` gradients too). So the picker swatch won't match what renders. The old `editorClasses: bg-[#222222]` existed to fix exactly this and got dropped.

Worth a short comment here so the next person doesn't re-break it:

```js
// Preview hexes are hardcoded: the Tina editor runs its own Tailwind instance
// without this repo's gray overrides (gray-950=#222222, gray-900=#333333), so
// bg-gray-* can't resolve here. Keep the dark hexes matching the render classes.
```

🤖

components/layout/homeTheme.tsx
```diff
@@ -0,0 +1,151 @@
...
+  if (typeof window === "undefined") return DEFAULT_THEME;
+  try {
+    const stored = window.localStorage.getItem(STORAGE_KEY);
+    if (stored === "light" || stored === "dark") return stored;
+  } catch {
+    /* ignore */
+  }
+  return window.matchMedia("(prefers-color-scheme: dark)").matches
+    ? "dark"
+    : "light";
```
isaaclombardssw, line 60: The PR description says the homepage "defaults to dark", but this falls back to `prefers-color-scheme`, so a light-OS visitor with nothing stored gets the light homepage by default. Code and description disagree - looks like this half may not have been finished.

Not my call which way it should go, but the two should be consistent. If dark-first was the intent, this is the change:

```suggestion
  return DEFAULT_THEME;
```

If OS-preference is intentional, leave it and update the PR description instead. Flagging so it's a deliberate choice, not an oversight - especially while light mode is still unreviewed.

🤖

components/layout/homeNavActions.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  const submit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    if (term.trim()) {
+      window.open(
+        `https://www.google.com.au/search?q=site:${SEARCH_SITE}%20${encodeURIComponent(
+          term
+        )}`,
+        "_blank"
+      );
```
isaaclombardssw, line 71: `window.open(..., "_blank")` without `noopener` lets the opened tab reach back via `window.opener` (reverse tabnabbing). Add the window features:

```suggestion
      window.open(
        `https://www.google.com.au/search?q=site:${SEARCH_SITE}%20${encodeURIComponent(
          term
        )}`,
        "_blank",
        "noopener,noreferrer"
      );
```

🤖

components/layout/homeNavActions.tsx
```diff
@@ -0,0 +1,209 @@
...
+  if (!mounted) return null;
+  return createPortal(
+    <div className={isDark ? "dark" : undefined}>{children}</div>,
+    document.body
+  );
+}
+
+function SearchButton({ isDark }: { isDark: boolean }) {
+  const [open, setOpen] = useState(false);
+  const [term, setTerm] = useState("");
```
isaaclombardssw, line 53: The search overlay can't be dismissed with the keyboard - Esc does nothing. Add a listener while it's open (`useEffect` is already imported):

```suggestion
  const [open, setOpen] = useState(false);
  const [term, setTerm] = useState("");

  useEffect(() => {
    if (!open) return;
    const onKey = (e: KeyboardEvent) =>
      e.key === "Escape" && setOpen(false);
    document.addEventListener("keydown", onKey);
    return () => document.removeEventListener("keydown", onKey);
  }, [open]);
```

Same gap on `RegionGlobe` below - worth the same treatment.

🤖

components/layout/homeNavActions.tsx
```diff
@@ -0,0 +1,209 @@
...
+        className={ICON_BUTTON}
+      >
+        <SearchIcon className="size-6" aria-hidden="true" />
+      </button>
+      {open && (
+        <Portal isDark={isDark}>
+          <div
+            className="fixed inset-0 z-1000 flex items-start justify-center bg-black/50 p-6 pt-24"
+            onClick={() => setOpen(false)}
+          >
```
isaaclombardssw, line 95: The modal has no dialog semantics, so screen readers don't announce it as a dialog or scope to it. Add the roles:

```suggestion
          <div
            role="dialog"
            aria-modal="true"
            aria-label="Site search"
            className="fixed inset-0 z-1000 flex items-start justify-center bg-black/50 p-6 pt-24"
            onClick={() => setOpen(false)}
          >
```

`autoFocus` on the input already handles focus-in; a full focus-trap (keeping Tab inside the dialog) is the one remaining a11y gap, but that's more than a basic and fine to leave for a follow-up.

🤖

isaaclombardssw (changes requested):
A few small bits of feedback.

Page looks good.

isaaclombardssw:
Resolve conflicts

isaaclombardssw (approved):
✅

Will eventually want dark mode across sites so that the mega menu isn't so brittle but this is fine for now

PR #4869, opened 2026-07-14
💄 Make homepage people cards clickable + fix console errors

wicksipedia:
## Why

The people cards on the homepage ("We work together to form an amazing collective brain") weren't clickable — you had to hit the small SSW logo icon to reach someone's profile. Making the whole card a link surfaced a pre-existing hydration error that was killing the block, and chasing that turned up a handful of other console/build warnings worth clearing while we were in here.

## What changed

Each fix is a separate commit.

### The feature

- **Clicking anywhere on a person's card opens their SSW People profile.** The card-wide link is an overlay rather than a wrapper, so the LinkedIn, X and SSW People icons stay separately clickable instead of nesting anchors inside one another. Cards with no SSW People URL stay non-clickable, and icons already hid themselves when their URL was missing.

### The bug that broke the clicks

- **Hydration mismatch in the person card textures.** `personCardTexture.tsx` used `useId()` to namespace its SVG `<defs>`; server and client generated different ids, so React bailed out of hydrating the whole block and the card went dead. The texture is already chosen deterministically from the card's index, so the id is now derived the same way. A `scope` keeps the mobile-carousel and desktop-grid copies unique.

### Console + build noise

- **`url(null)` → `GET /null`.** `v2ComponentWrapper` always set `background-image`, so any block without one rendered `url(null)`. The browser dutifully fetched `/null`, which hit the `[filename]` catch-all and logged `Unable to find record content/pages/null.mdx` on every page load. One guard, both symptoms gone.
- **`collapsible` leaking to the DOM.** Radix's `Accordion` only accepts `collapsible` with `type="single"`. With `type="multiple"` it fell through to the DOM and React logged ``Received `true` for a non-boolean attribute `collapsible` `` on every page using the block.
- **Legacy `objectFit` prop** on both logo carousels — dropped by `next/image` in v13, so it did nothing and warned on every logo. Replaced with `object-contain`.
- **Undeclared image quality.** The hero and event banners render at `quality={100}`, which Next 16 will refuse to serve unless it's listed. Added `images.qualities` to `next.config.mjs`.
- **Ambiguous `z-[2]`.** The arbitrary value matched both the core `z` utility and the `zIndexStack` plugin's, so Tailwind warned and the class was unreliable. Added `2` to the `zIndex` scale; the hero uses the named `z-2`.
- **Stale browserslist data** (14 months old). Ran `update-browserslist-db`, which also deduped `caniuse-lite` from three copies to one and patch-bumped two transitive deps (`crelt`, `uuid`). No target browser changes.

## Verification

Driven in a real browser against a dev server:

- Clicking the photo navigates to `ssw.com.au/people/adam-cogan/`; clicking the LinkedIn icon on the same card navigates to `linkedin.com/in/adamcogan` — the overlay and the icons each win their own clicks.
- Homepage: zero console errors, zero `/null` requests, no duplicate texture ids in the DOM.
- Consulting page (which uses the accordion block): the `collapsible` error is gone.
- `tsc --noEmit` clean.

## Not fixed (out of scope, flagging them)

- The **"More SSW People"** button 404s in dev. `/people` is a separate app served at `ssw.com.au/people` in production — nothing in this repo serves that route, so this is expected locally and fine in prod.
- `/api/tina/query/with-fallbacks` 404s on the consulting page. That's Tina's own dev-only endpoint, unrelated to anything here.


wicksipedia:
Good catch — the comment read as though the social links were `z-10`, but that's the overlay; the icons are `z-20`. Reworded in ad2f801. No behaviour change.

joshbermanssw:
Border effect still on whole card during hover

isaaclombardssw (approved):
✅✅✅ good improvements

PR #4884, opened 2026-07-22
⚡ Homepage performance, accessibility, and live-stream fixes

joshbermanssw:
## TL;DR

Works through the homepage's Lighthouse findings — a heavy LCP hero image, a 0.698 CLS footer jump, and a forced reflow — and folds in a footer-contrast fix, a testimonial ARIA fix, and a site-wide live-stream hydration error. No visual or behavioural change is intended; the perf items map 1:1 to the flagged Lighthouse audits.

## What's here

Three independent concerns on one branch:

**⚡ Page-speed (Lighthouse)**
- **LCP / image weight** — hero background dropped from `quality={100}` to `75` (imperceptible under its gradient overlay + 60% opacity) and a `1440` device size closes the `1200 → 1920` srcset gap, so ~1250–1310px slots stop pulling the oversized 1920 candidate. Hero est. ~126 KiB → ~40 KiB.
- **CLS (0.698)** — the globe stage (34–40rem tall) was gated on a client-only media query, so it was absent from the SSR HTML and only appeared after hydration, shoving the footer (the last block on the page) down ~a full viewport. The stage box is now reserved via CSS (`hidden lg:block`) with only the WebGL canvas waiting on `isDesktop`; also drops a redundant `h-screen` from the page wrapper.
- **Forced reflow** — `ButtonRow` read each button's `clientWidth` in a post-render effect (a synchronous layout, once per ButtonRow). The reads now run inside the `ResizeObserver` callback, which fires after layout is already computed.

**♿ Accessibility**
- Testimonials: prohibited `aria-label` on a generic span replaced with an sr-only quote copy + `aria-hidden` animated words.
- Footer "powered by" row: `text-gray-500` (3.33:1) → `text-gray-400` (6.34:1) to clear WCAG AA contrast.

**🐛 Live-stream banner**
- Mount-gate the time-dependent banner so cached SSR HTML and the first client render match, resolving the site-wide React #418 hydration error.
- Derive banner timezone/labels from the streamed chapter's city (Sydney stays as fallback), show "Streaming has finished." for past events, and clamp `countdownTextFormat` against negative input.

## Files

| File | Concern |
|---|---|
| `components/blocks/v3/heroBox/heroBox.tsx`, `next.config.mjs` | LCP / image weight |
| `components/blocks/v3/globe/globe.tsx`, `app/components/page-layout.tsx` | CLS |
| `components/blocksSubtemplates/buttonRow.tsx` | Forced reflow |
| `components/blocks/v3/testimonials/testimonials.tsx`, `components/layout/footer/footer.tsx` | A11y |
| `app/live-steam-banner/live-stream.tsx`, `components/liveStream/liveStreamBanner.tsx`, `helpers/countdownTextFormat.ts` | Live-stream |

## Reviewer notes

- **Perf items map 1:1 to Lighthouse.** The image "larger than needed" flag, the 0.698 footer CLS, and the forced-reflow audit — re-run Lighthouse on the homepage to confirm.
- **Button wrap behaviour unchanged.** When one button fills the row, the others still stack full-width; the measurement just moved off the render path.

### DIFF
--- a/app/components/page-layout.tsx
+++ b/app/components/page-layout.tsx
@@ -13,7 +13,7 @@ const PageLayout = ({
 }: PageLayoutProps) => {
   return (
     <HeaderAppearanceProvider>
-      <div className="flex h-screen min-h-screen flex-col">
+      <div className="flex min-h-screen flex-col">
         <header className="no-print z-1">
           {phishingBanner}
           {megaMenu}
--- a/app/live-steam-banner/live-stream.tsx
+++ b/app/live-steam-banner/live-stream.tsx
@@ -52,6 +52,12 @@ interface LiveStreamProps extends PropsWithChildren {
 export function LiveStream({ event, children }: LiveStreamProps) {
   const [countdownMins, setCountdownMins] = useState<number>();
   const [liveStreamDelayMinutes, setLiveStreamDelayMinutes] = useState(0);
+  // Gate all time-dependent output behind mount so the server render and the
+  // first client render are identical. The root layout is cached (revalidate),
+  // so a render-time clock check would otherwise disagree between the cached
+  // server HTML and the live client, causing a hydration mismatch (React #418).
+  const [mounted, setMounted] = useState(false);
+  useEffect(() => setMounted(true), []);
 
   const eventDynamic: EventInfo = {
     ...event,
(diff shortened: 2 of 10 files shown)


isaaclombardssw (approved):
✅

PR #4887, opened 2026-07-22
⚡ Import icons by name to drop 4.76 MB from the bundle

joshbermanssw:
## TL;DR

`icon.tsx` shipped the entire ~3,300-icon Simple Icons set — **4.76 MB** — to every page, in order to render the single icon used on `/consulting/bots`. Replacing three `import * as` namespace imports with named imports of the 104 icons content actually references drops the largest chunk from **4.76 MB to 1.55 MB**.

## Why

Three compounding issues meant nothing could tree-shake:

```js
import * as SimpleIcons from "react-icons/si";                                  // 1. namespace import
export const IconOptions = { Tina, ...BoxIcons, ...SimpleIcons, ...CodeIcons }; // 2. spread
const IconSVG = IconOptions[name];                                              // 3. dynamic lookup
```

`import * as` opts out of named-import tree-shaking, the spread materialises every export into a runtime object, and the lookup is by a runtime string — so the bundler can never prove an icon is unused and has to keep all of them.

Actual usage across all of `content/`:

| Set | Used | Shipped |
|---|---|---|
| Si | **1** (`SiAudiobookshelf`) | entire set, 4.76 MB |
| Bi | 101 | entire set |
| Vsc | 2 | entire set |

## Verification

Production build (`next build`, exit 0):

- Largest chunk **4.76 MB → 1.55 MB**; no Simple Icons remain in `.next/static/chunks`.
- All 104 used icons still present and resolving (spot-checked `BiAbacus`, `BiCheckShield`, `SiAudiobookshelf`, `VscDashboard`, `VscCommentDiscussion`).
- All 104 names validated against the packages' declared exports, so a mistyped name fails the build rather than silently rendering nothing.

## Reviewer notes

- **The picker catalogue is now 104 icons.** `iconSelector` derives its list from `Object.keys(IconOptions)`, so editors browse only icons already in use; adding a new one means adding its named import. Deliberate call — only 3 non-Bi icons have ever been used across the whole site.
- **No visual change.** Everything currently referenced in content renders identically.
- **Remaining large chunks are out of scope.** The 1.55/1.14/1.09 MB chunks are the TinaCMS + Slate + CodeMirror editor stack reaching public pages via `useTina`; that's a separate piece of work.


isaaclombardssw (approved):
✅

PR #4910, opened 2026-07-23
💸 Perf - Defer GTM to lazyOnload without dropping early events

### LINKED ISSUE #4908: 💸 Perf - Defer GTM and its marketing tags out of the critical window

**Description**

`app/layout.tsx` loads Google Tag Manager via `<GoogleTagManager>` from `@next/third-parties/google` (`afterInteractive` strategy). GTM then pulls in the marketing/analytics tags configured in the container — on the PageSpeed run these were GA4 (`gtag`), Google Ads (`AW-…`), the Facebook Pixel (`fbevents.js`) and Hotjar — roughly 650 KB combined, landing in the hydration window and adding to TBT and JS-execution time.

None of these need to run before the page is interactive.

**Proposed solution**

1. Load GTM with `lazyOnload` instead of `afterInteractive` (on idle, after the page is usable). `@next/third-parties`' `GoogleTagManager` doesn't expose a strategy prop, so this likely means the standard GTM snippet in a plain `<Script strategy="lazyOnload">`.
2. Keep `window.dataLayer` defined eagerly so early `dataLayer.push` / `sendGTMEvent` calls queue instead of being dropped before GTM loads — e.g. the `eventbrite_order_complete` conversion in `components/eventbrite/eventbriteModalButton.tsx`.
3. Where most of the weight actually lives — the container: review the triggers on the heavy tags (GA / Ads / FB / Hotjar) so they fire on Window Loaded / interaction / a short timer rather than the initial page view. This is GTM config, not code, and is where the bulk of the 650 KB moves off the critical path.

**Verification — the tags must still fire**

- **Network:** confirm each beacon still goes out after load — GA4 `google-analytics.com/g/collect`, Ads `googleads.g.doubleclick.net/pagead`, FB `facebook.com/tr?`, Hotjar `*.hotjar.io`.
- **Authoritative:** GTM Tag Assistant (which tags fired + order), GA4 Realtime, Meta Events Manager → Test Events, Hotjar dashboard.
- **Regression:** a Playwright test that loads the page, waits for idle, and asserts the beacons fire — so a later refactor can't silently break tracking.
- **Edge case:** click a tracked CTA immediately after load and confirm the conversion still records, not just the delayed page view.

**Acceptance criteria**

1. GTM and its downstream tags no longer execute before the page is interactive
2. TBT / JS-execution measurably improves on the events page
3. GA4 page views, Google Ads + Eventbrite conversions, the Facebook Pixel and Hotjar all still fire (verified by the checks above), including a conversion triggered immediately after load
4. A regression test asserts the marketing beacons fire

**Note**

GTM is loaded site-wide from the root layout, so this applies to every page — it also overlaps the homepage PBI #4889, not just this event page.

isaaclombardssw:
Closes #4908.

## What changed

`app/layout.tsx` loaded GTM via `<GoogleTagManager>` from `@next/third-parties/google`, which only offers an `afterInteractive` load — so GTM (and the GA4/Ads/FB/Hotjar tags configured in the container) landed in the hydration window.

Replaced it with a small local `GoogleTagManager` component (`app/components/google-tag-manager.tsx`) that emits the equivalent GTM snippet as two `next/script` tags:

- **eager init** (`afterInteractive`): `window.dataLayer = window.dataLayer || []` + the `gtm.start` push, so the dataLayer exists and early events queue.
- **`gtm.js` itself** on `strategy="lazyOnload"` (loads on idle, after the page is interactive).

The call site is unchanged: `<GoogleTagManager gtmId={process.env.NEXT_PUBLIC_GOOGLE_GTM_ID} />`. Added a guard so nothing renders when the id is unset (previously it would have requested `gtm.js?id=undefined`).

## Conversions / dataLayer still work

`sendGTMEvent` from `@next/third-parties/google` does **not** depend on `<GoogleTagManager>` rendering — its source pushes straight to `window.dataLayer` (defaulting the layer name to `dataLayer` and creating the array if missing):

```js
const sendGTMEvent = (data, dataLayerName) => {
  const dataLayer = dataLayerName || currDataLayerName; // currDataLayerName defaults to "dataLayer"
  window[dataLayer] = window[dataLayer] || [];
  window[dataLayer].push(data);
};
```

So the `eventbrite_order_complete` conversion in `components/eventbrite/eventbriteModalButton.tsx` still queues into `window.dataLayer` and is flushed when GTM finishes loading on idle — including the edge case of a CTA clicked immediately after load, since the push is queued, not dropped.

## Regression test

`ui-tests/gtm-tracking.spec.ts` (Playwright, existing `ui-tests/` convention) loads `/`, waits for network idle, then asserts:

1. `gtm.js` is still requested (tracking isn't silently disabled by the deferral), and
2. `window.dataLayer` contains the `gtm.js` start event (queuing is intact).

It self-skips the live-beacon assertion when `NEXT_PUBLIC_GOOGLE_GTM_ID` isn't set on the target server (e.g. local dev), so it's meaningful against a real environment and green everywhere else. Run it against a server that has the id, e.g. `HOST_URL=https://www.ssw.com.au pnpm exec playwright test ui-tests/gtm-tracking.spec.ts`.

## Out of code scope — follow-up for whoever owns the GTM container

The bulk of the ~650 KB win is **retiming the individual tag triggers** (GA4 / Google Ads / Facebook Pixel / Hotjar) to fire on Window Loaded / interaction / a short timer instead of the initial page view. That's GTM container config, not code, and lives entirely in the container. This PR only changes *how/when GTM itself loads*.

## Manual dashboard verification (needs the real container)

- **GTM Tag Assistant** — which tags fired and in what order
- **GA4 Realtime** — page views land
- **Meta Events Manager → Test Events** — Pixel events land
- **Hotjar dashboard** — session recorded
- Network beacons: GA4 `google-analytics.com/g/collect`, Ads `googleads.g.doubleclick.net/pagead`, FB `facebook.com/tr?`, Hotjar `*.hotjar.io`


isaaclombardssw:
Needs GTM follow up

isaaclombardssw:
Closed as a duplicate of Josh's fix #4933

Issue #4894, opened 2026-07-22
💸 Perf - AI for Business Leaders event page - LCP 9.3s, TBT 1,250ms on mobile


joshbermanssw:
## Pain

Mobile Lighthouse 13.4 on https://www.ssw.com.au/events/ai-for-business-leaders (Moto G Power, Slow 4G, 23 Jul 2026) is poor:

| Metric | Value | Score contribution |
|---|---|---|
| First Contentful Paint | 2.8 s | +6 |
| Largest Contentful Paint | **9.3 s** | **+0** |
| Total Blocking Time | **1,250 ms** | **+6** |
| Cumulative Layout Shift | 0.133 | +20 |
| Speed Index | 4.1 s | +8 |

Accessibility is 81. Total page weight 2,867 KiB. Lighthouse also flags: unused JS 1,064 KiB, JS execution 2.9 s, main-thread work 3.9 s, 17 long tasks, forced reflow, inefficient cache lifetimes (246 KiB), legacy JavaScript (41 KiB).

## TL;DR of the root causes

The page itself is simple — 39 blocks of only 6 component types (20 × `spacer`, 8 × `imageTextBlock`, 4 × `cardCarousel`, 2 × `accordionBlock`, 1 × `logoCarousel`, 1 × `breadcrumbs`). The cost is almost entirely framework and shared-component overhead, not content:

1. **1.1 MB of TinaCMS's MDX parser ships in the initial bundle of every page.** From `.next/app-build-manifest.json`, this route's entry is 1.70 MB of JS, and `dd3583a7-*.js` alone is 1,115 KB of it. 65 components import `TinaMarkdown` from `tinacms/dist/rich-text`, whose first line is `import { sanitizeUrl } from "@tinacms/mdx"` — a 1.97 MB package (remark/mdast/slate/acorn) pulled in for one URL sanitiser. This accounts for essentially all of Lighthouse's "unused JavaScript — 1,064 KiB".
2. **The LCP image is lazy-loaded.** `blocks-renderer.tsx` grants `priority` to block 0, which is `breadcrumbs` (no image). The hero is block 2 and gets `loading="lazy"` with no `fetchpriority`. The file is only 78 KB — 9.3 s is entirely "requested late, behind 1.7 MB of JS".
[...]

PR #4911, opened 2026-07-23
🐛 Resolve mobile header appearance server-side to fix CLS

### LINKED ISSUE #4906: 💄 Header - Mobile header flashes flag/Contact for one frame before hiding
Sub-issue of the AI for Business Leaders performance PBI. Highest blast radius of the set — schedule it last.

**Description**

CLS is 0.133 and the cause is already documented in the code. `app/components/header-appearance.tsx` carries this note:

> client context → SSR shows full header for one frame before hiding flag/contact. Upgrade path if the flash matters: set `x-pathname` in middleware.ts and resolve appearance server-side in app/layout.tsx so the header renders correctly at SSR.

The header server-renders with the country flag and the Contact button present. After hydration, `useMobileHeaderAppearance` pushes the page's `appearance` into context and both are removed, reflowing the header and everything below it.

This page sets **both** flags in `content/eventsv2/ai-for-business-leaders.json`:

```json
"appearance": { "hideFlag": true, "hideContactButton": true }
```

so it takes the full shift.

**Proposed solution**

Follow the upgrade path already written in the comment:

1. Set an `x-pathname` header in `middleware.ts`
2. Resolve the page's `appearance` server-side in `app/layout.tsx` and render the header correctly on the first paint
3. Keep the client context as the mechanism for client-side navigations

Because this touches middleware and the root layout, it affects every route — test broadly.

**Acceptance criteria**

1. The header renders in its final form on first paint, with no post-hydration flag/Contact removal
2. CLS on `/events/ai-for-business-leaders` is below 0.1
3. Client-side navigation between a page that hides the flag and one that doesn't still updates the header correctly
4. The homepage, consulting pages and event pages all render the correct header variant
5. No hydration mismatch warnings

isaaclombardssw:
Closes #4906 (parent PBI #4894, item 12).

## Root cause

The header (`MegaMenuWrapper`) renders in the root layout, which has no access to the current page's data. So the per-page decision to hide the mobile country flag / Contact button ran in a client `useEffect` (`useMobileHeaderAppearance`) that pushed the page's `appearance` into React context **after hydration**. The header then removed those elements, reflowing itself and everything below it — CLS 0.133 on `/events/ai-for-business-leaders`, which sets both `hideFlag` and `hideContactButton`.

## Fix

Render the final state server-side and let CSS do the hiding, so the first paint is already correct:

- An opting-out page server-renders a hidden marker (`HeaderAppearanceMarker`) carrying `data-mm-hide-flag` / `data-mm-hide-contact`.
- Unlayered CSS in `styles.css` reads it with `body:has([data-mm-hide-flag]) …` and hides the matching `ssw.megamenu` elements. No header-side JS.
- `MegaMenuWrapper` now always renders the flag/Contact button (it only still hides the phone on `/company/contact-us`, which it already knows at SSR from the pathname).
- The client context/provider/hook are removed — CSS covers both SSR **and** client-side navigation (React swaps the marker in/out with the page).

## Why CSS, not the middleware path the comment documented

`header-appearance.tsx` previously suggested an `x-pathname` middleware header + server-side resolution in the layout. I did **not** take that route: reading `headers()` in the **root** layout opts every route out of static generation, and directly conflicts with `export const dynamic = "force-static"` on the events route — it would break the static build for a heavily-cached marketing site. The CSS approach keeps every route static and adds zero runtime cost.

Trade-off to be aware of: the CSS selectors target `ssw.megamenu@4.13.8` internals (the flag `img[alt$="flag"]`, the mobile divider, the `.gap-2.sm:flex-grow-0` phone-button wrapper). The version is pinned; there's a comment to revisit on upgrade. Cleaner long-term would be a stable `data-*` hook or className prop upstreamed into `ssw.megamenu`.

## Expected impact

CLS on `/events/ai-for-business-leaders` should drop from 0.133 toward ~0 for the header contribution, since the flag/Contact button no longer appear-then-disappear.

## Verified

- `pnpm lint` passes (only the pre-existing, unrelated `z-[2]` tailwind warning).
- Prettier clean on all changed files.
- CSS diff kept minimal (existing lines left untouched).

## Still needs manual verification

- `pnpm build` / typecheck could **not** be run here — Tina codegen requires TinaCloud creds (`clientId`/`token`/`branch`), so the generated client isn't present. Please run a full build in CI/locally with creds.
- Re-run mobile Lighthouse on `/events/ai-for-business-leaders` to confirm **CLS < 0.1**.
- Visually check mobile + desktop headers on: homepage, a consulting page, a normal event page (flag + Contact present), and `/events/ai-for-business-leaders` (both hidden, no post-hydration shift). Confirm client-side nav between a hiding page and a non-hiding page updates the header correctly, and no hydration-mismatch warnings.

### DIFF
--- a/app/(events)/events/[...filename]/eventsv2.tsx
+++ b/app/(events)/events/[...filename]/eventsv2.tsx
@@ -1,5 +1,5 @@
 "use client";
-import { useMobileHeaderAppearance } from "@/app/components/header-appearance";
+import { HeaderAppearanceMarker } from "@/app/components/header-appearance";
 import { Blocks } from "@/components/blocks-renderer";
 import { Container } from "@/components/util/container";
 import { Section } from "@/components/util/section";
@@ -17,9 +17,9 @@ type EventsV2PageProps<T> = {
 const EventsV2Page = memo(
   function EventsV2Page({ tinaProps }: EventsV2PageProps<object>) {
     const { blocks, appearance } = tinaProps.data.eventsv2;
-    useMobileHeaderAppearance(appearance ?? null);
     return (
       <div className="dark flex h-full flex-col">
+        <HeaderAppearanceMarker appearance={appearance} />
         <Section color={"toggleLightMode"}>
           <Container
             size="custom"
--- a/app/components/header-appearance.tsx
+++ b/app/components/header-appearance.tsx
@@ -1,44 +1,27 @@
-"use client";
-import { createContext, useContext, useEffect, useMemo, useState } from "react";
-
 export type MobileHeaderAppearance = {
   hideFlag?: boolean | null;
   hideContactButton?: boolean | null;
 };
-type Ctx = {
-  mobile: MobileHeaderAppearance;
-  setMobile: (v: MobileHeaderAppearance) => void;
-};
-const HeaderAppearanceContext = createContext<Ctx>({
-  mobile: {},
-  setMobile: () => {},
-});
 
-export function HeaderAppearanceProvider({
-  children,
+// The header lives in the root layout and can't see the current page's data, so
+// hiding the mobile flag/Contact button used to happen in a client effect AFTER
+// hydration — reflowing the header and everything below it (CLS). Instead, the page
+// server-renders this hidden marker; CSS in styles.css reads it via `:has()` and hides
+// the elements on first paint, so SSR and hydration match. Client-side navigation swaps
+// the marker in/out with the page, so the header updates without any header-side JS.
+// (Middleware `x-pathname` + server resolution — the other documented fix — was rejected:
+// reading headers() in the root layout breaks the `dynamic = "force-static"` routes.)
+export function HeaderAppearanceMarker({
+  appearance,
 }: {
-  children: React.ReactNode;
+  appearance?: MobileHeaderAppearance | null;
 }) {
-  const [mobile, setMobile] = useState<MobileHeaderAppearance>({});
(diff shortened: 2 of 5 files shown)


isaaclombardssw:
## Measured: CLS on mobile

Staging slot (this branch) vs prod (main), `/events/ai-for-business-leaders`, mobile Lighthouse, median of 3 runs:

| Metric | Before (prod) | After (this PR) | Δ |
|---|---:|---:|---:|
| **CLS** | **0.133** | **0.008** | **-94%** |
| Performance score | 69 | 81 | +12 |
| LCP | 3.7 s | 2.2 s | (see note) |
| TBT | 618 ms | 672 ms | ~noise |

CLS is the clean, attributable win: the header no longer strips the flag/Contact button after hydration, so the content doesn't shift. 0.008 is comfortably under the 0.1 "good" threshold (AC #3). The perf-score bump mostly follows the CLS/LCP change; TBT/LCP swings are partly run-to-run noise and slot-vs-prod variance, so I'm anchoring on CLS.

**Repro:**
```bash
npx lighthouse@12 https://app-sswwebsite-9eb3-pr-4911.azurewebsites.net/events/ai-for-business-leaders \
  --form-factor=mobile --screenEmulation.mobile --only-categories=performance \
  --quiet --chrome-flags="--headless=new" --output=json --output-path=after.json
# baseline: same command against https://www.ssw.com.au/events/ai-for-business-leaders
```
Baseline = prod (main). This branch is a few commits off main, so it isn't a fully isolated single-variable build, but CLS on this page is driven by the header change.

Issue #4892, opened 2026-07-22
⚡ Keep the TinaCMS editor stack (~3.8 MB) off public pages


joshbermanssw:
Parent: #4889 · **Priority 3** — largest remaining payload, but a genuine refactor with real risk to the editing experience.

## Problem

The production homepage references 61 chunks, including the TinaCMS editor stack:

| Chunk | Size | Contents |
|---|---|---|
| `5994` | 1.56 MB | Slate + yjs |
| `41958015` | 1.12 MB | TinaCMS + Slate + highlight.js |
| `dd3583a7` | 1.08 MB | TinaCMS admin + CodeMirror |

These are admin-editor dependencies being parsed and evaluated by ordinary visitors. They don't appear in the route's *static* manifest because they arrive via `dynamic()` imports, which Next still preloads into the HTML.

## Root cause

**9 files value-import the full `tinacms` package** for admin form-field helpers (`wrapFieldsWithMeta`, `Button`, `Input`, `TextArea`):

- `components/blocks/tableLayout.schema.tsx`
- `components/blocks/cardCarousel/cardCarousel/cardCarouselSchema.tsx`
- `components/blocks/v3/cardCarousel/cardCarousel.schema.tsx`
- `components/blocks/logoCarousel/logoCarouselSchema.tsx`
- `components/blocks/breadcrumbs/breadcrumbs.schema.tsx`
- `components/fields/color.tsx`
- `components/blocksSubtemplates/tinaFormElements/iconSelector.tsx`
- `components/blocksSubtemplates/tinaFormElements/colourSelector.tsx`
- `components/textInputWithCount/textInputWithCount.tsx`

The other 92 files import from `tinacms/dist/react`, which is the lighter runtime and fine. Note the ~40 `import type { Template } from "tinacms"` are type-only and erased at compile time — harmless.

The problem is that schema files are reachable from the runtime component graph, so admin-only code follows the blocks onto public pages.

## Suggested approach

Establish a hard boundary between **schema/admin** code and **runtime/public** code so the full `tinacms` package is never statically reachable from a rendered block — e.g. lazy-load field components behind `dynamic()` split points, or move field UI into an admin-only module.

## Watch out for

This touches the CMS editing experience. Needs proper scoping and verification that every Tina form field still renders and saves.

joshbermanssw:
[...]

PR #4912, opened 2026-07-23
⚡ Alias @tinacms/mdx to keep its 2 MB markdown parser out of the client bundle

### LINKED ISSUE #4896: 💸 Perf - Keep the 1.1 MB TinaCMS MDX parser out of the client bundle
Sub-issue of the AI for Business Leaders performance PBI. **This is the single biggest lever.**

**Description**

The events route ships **1.70 MB of JS** in its initial entry (`.next/app-build-manifest.json` → `/(events)/events/[...filename]/page`). One chunk, `dd3583a7-*.js`, is **1,115 KB** of that — roughly 65%. Fingerprinting it shows remark/mdast/slate internals (`invalid_markdown`, `Error parsing markdown`).

The chain:

- 65 components across `app/` and `components/` do `import { TinaMarkdown } from "tinacms/dist/rich-text"`
- Line 1 of `tinacms/dist/rich-text/index.js` is `import { sanitizeUrl } from "@tinacms/mdx"`
- `@tinacms/mdx/dist/index.browser.js` is **1.97 MB** — the full MDX parser and serialiser

So the entire MDX toolchain is bundled for anonymous visitors in order to call one ~15-line URL sanitiser. This is almost exactly Lighthouse's "Reduce unused JavaScript — Est savings of 1,064 KiB", and it is the main driver of "Reduce JavaScript execution time 2.9 s" and TBT 1,250 ms.

Worth noting what is *not* the problem: `tinacms/dist/react.js` (where `useTina` and `tinaField` come from) is only 6 KB and imports from `@tinacms/bridge`, not from the 2.5 MB `tinacms/dist/index.js`. Those imports are fine and don't need changing.

**Proposed solution**

Option A (preferred — small, reversible): alias `@tinacms/mdx` to a local module exporting a faithful `sanitizeUrl`. It must be applied in **both** bundlers — `webpack.resolve.alias` in `next.config.mjs` for `next build`, and `turbopack.resolveAlias` for `next dev --turbopack`. No app code imports `@tinacms/mdx` directly, so the blast radius is limited to what `tinacms/dist/rich-text` uses.

Option B: vendor `TinaMarkdown` locally (12 KB of MIT-licensed source) as e.g. `@/components/tinaMarkdown`, implement `sanitizeUrl` alongside it, and repoint the 65 imports with a codemod.

Option C (longer term, in parallel): raise it upstream — `sanitizeUrl` belongs in `@tinacms/bridge`, not behind the whole MDX package.

**Acceptance criteria**

1. `dd3583a7-*.js` (or its successor) no longer appears in the events route's entry in `app-build-manifest.json`
2. The route's initial JS drops from 1.70 MB to under 700 KB uncompressed
3. Rich text still renders identically on `/events/ai-for-business-leaders` and at least two other pages using `TinaMarkdown` (an `articles/` page and a `consulting/` page)
4. TinaCMS editing (`/admin`) and visual editing still work
5. `next dev` (Turbopack) and `next build` (webpack) both produce correct output
6. Lighthouse "Reduce unused JavaScript" drops by roughly 1 MB

joshbermanssw:
- Affected routes: all routes rendering rich text (every page importing `TinaMarkdown` — 65 components), verified on `/events/ai-for-business-leaders`

- Fixed #4896

## TL;DR

`tinacms/dist/rich-text` — the module behind `<TinaMarkdown>` — opens with `import { sanitizeUrl } from "@tinacms/mdx"`, and that package ships a single `bundleDeps` entry point of **1,973,803 bytes** to supply a **22-line** URL sanitiser. This aliases the package to a shim that keeps `sanitizeUrl` and drops the parser; no application code changes.

## Why

On the deployed site, `/events/ai-for-business-leaders` references **5,966,858 bytes of JS across 60 chunks**. `dd3583a7-*.js` — the `@tinacms/mdx` bundle — is **1,142,097** of them, making it the largest single attributable slice of the page's JavaScript and the bulk of Lighthouse's "Reduce unused JavaScript — Est savings of 1,064 KiB".

Reproduce the page total:

```bash
curl -s https://www.ssw.com.au/events/ai-for-business-leaders \
  | grep -o '/_next/static/chunks/[A-Za-z0-9_.-]*\.js' | sort -u \
  | while read c; do curl -s -o /dev/null -w "%{size_download}\n" \
      -H 'Accept-Encoding: identity' "https://www.ssw.com.au$c"; done \
  | awk '{s+=$1} END {print s" bytes"}'
```

Local before/after build figures to follow in a comment.

## Reviewer notes

- **`sanitizeUrl` is copied, not reimplemented.** Verbatim from `@tinacms/mdx@2.1.9` and byte-identical across that package's node and browser builds. If you bump the Tina packages, re-diff it.
- **`parseMDX` / `serializeMDX` throw.** Neither runs in a browser on a public page — `serializeMDX` is reached only from the editor's save path, `parseMDX` only from the Tina GraphQL layer, and both `/admin` and `@tinacms/cli` are built outside this Next config, so both still get the real package. They throw rather than returning a plausible value so a wrong assumption fails loudly instead of silently mangling content.
- **Aliased for both bundlers.** `next dev` uses Turbopack and `next build` uses webpack; wiring only one would let dev and prod diverge silently.
- **This is a stopgap.** tinacms/tinacms#7233 adds a dependency-free `@tinacms/mdx/sanitize-url` subpath that makes the shim unnecessary — delete the file and the two aliases once it ships in a release we consume.

## Tests

- Existing suite green (13/13).
- Differential test of the shim's `sanitizeUrl` against the real export across 31 inputs — allowed and blocked schemes (`javascript:`, `data:`, `vbscript:`, `file:`, `ftp:`), relative paths, anchors, userinfo, explicit ports, and query/hash/trailing-slash combinations: **0 mismatches**, and an identical export surface.

## Links

- Closes #4896
- Parent PBI #4894
- Upstream issue: tinacms/tinacms#7232
- Upstream fix: tinacms/tinacms#7233
- Related, same payload via a different import chain: #4892

### DIFF
--- a/lib/tinacms-mdx-shim.js
+++ b/lib/tinacms-mdx-shim.js
@@ -1,0 +1,81 @@
+/**
+ * Stand-in for `@tinacms/mdx`, aliased in via `next.config.mjs`.
+ *
+ * WHY THIS EXISTS
+ * ---------------
+ * `tinacms/dist/rich-text` — the module behind `<TinaMarkdown>`, imported by 65
+ * components in this repo — starts with:
+ *
+ *     import { sanitizeUrl } from "@tinacms/mdx";
+ *
+ * `@tinacms/mdx` publishes a single "." entry point built with `bundleDeps`, so
+ * that one import drags the whole remark / mdast / micromark / prettier parsing
+ * toolchain into the bundle: 1,973,803 bytes (@tinacms/mdx 2.1.9) to reach a
+ * 22-line URL sanitiser. It was ~1.14 MB minified on the deployed site and the
+ * single largest attributable slice of the page's JavaScript.
+ *
+ * Aliasing the package here keeps `sanitizeUrl` — the only export any browser
+ * code path actually needs — and drops the parser.
+ *
+ * HOW TO REMOVE THIS
+ * ------------------
+ * Upstream tinacms#7232 tracks the same problem and tinacms#7233 adds a
+ * dependency-free `@tinacms/mdx/sanitize-url` subpath. Once that ships in a
+ * release we consume, delete this file and its two aliases in `next.config.mjs`
+ * and bump `tinacms` / `@tinacms/mdx`. See SSW.Website#4896.
+ *
+ * KEEPING IT HONEST
+ * -----------------
+ * `sanitizeUrl` below is a verbatim copy from @tinacms/mdx@2.1.9 (byte-identical
+ * in both `dist/index.js` and `dist/index.browser.js`). If you bump the Tina
+ * packages, re-diff it against the new build before assuming it still matches.
+ */
+
+// Verbatim from @tinacms/mdx@2.1.9 — do not "improve" it; it must match upstream.
+export const sanitizeUrl = (url) => {
+  const allowedSchemes = ["http", "https", "mailto", "tel", "xref"];
+  if (!url) return "";
+  let parsedUrl = null;
+  try {
+    parsedUrl = new URL(url);
+  } catch (error) {
+    return url;
+  }
+  const scheme = parsedUrl.protocol.slice(0, -1);
+  if (allowedSchemes && !allowedSchemes.includes(scheme)) {
+    // eslint-disable-next-line no-console -- upstream behaviour, kept as-is
+    console.warn(`Invalid URL scheme detected ${scheme}`);
+    return "";
+  }
+  if (parsedUrl.pathname === "/") {
+    if (url.endsWith("/")) {
+      return parsedUrl.href;
+    }
+    return `${parsedUrl.origin}${parsedUrl.search}${parsedUrl.hash}`;
+  } else {
+    return parsedUrl.href;
+  }
+};
+
+/**
+ * The other two exports of `@tinacms/mdx`. Neither runs in a browser on a public
+ * page: `serializeMDX` is reached only from `tinacms/dist/index.js` on the
+ * editor's save path, and `parseMDX` only from the Tina GraphQL layer, which runs
+ * in the `@tinacms/cli` process and in `/admin` — both built outside this Next
+ * config, so both still get the real package.
+ *
+ * They throw rather than returning a plausible-looking value so that if that
+ * assumption is ever wrong we get a loud, traceable failure instead of silently
+ * mangled content.
+ */
+const unavailable = (name) => () => {
+  throw new Error(
+    `${name}() is not available: @tinacms/mdx is aliased to lib/tinacms-mdx-shim.js ` +
+      `to keep the ~2 MB markdown parser out of the browser bundle. If you genuinely ` +
+      `need ${name}() in the Next build, remove the alias in next.config.mjs and ` +
+      `re-measure the bundle. See SSW.Website#4896.`
+  );
+};
+
+export const parseMDX = unavailable("parseMDX");
+export const serializeMDX = unavailable("serializeMDX");
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -4,11 +4,16 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 const withNextPluginPreval = createNextPluginPreval();
 
-// Absolute so webpack and Turbopack resolve it identically regardless of cwd.
-const TINACMS_MDX_SHIM = path.join(
+// The two bundlers want this path in different forms, so don't unify them:
+// webpack's resolve.alias needs an absolute path, while Turbopack's resolveAlias
+// resolves values against the project root and prepends "./" to whatever it is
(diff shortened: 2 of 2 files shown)


joshbermanssw:
## Before/after measured

Two clean production builds on the same machine, identical apart from the alias (baseline = `next.config.mjs` from `origin/main` with the shim removed).

| | Before | After | Δ |
|---|---:|---:|---:|
| `/(events)/events/[...filename]` entry JS | 1,752,504 B | **610,909 B** | **−1,141,595 (−65.1%)** |
| Entry chunk count | 22 | 22 | — |
| All `static/chunks` bytes | 10,475,021 B | 9,346,839 B | −1,128,182 (−10.8%) |

`dd3583a7-76595befd5b43fbb.js` (1,142,097 B) — the `@tinacms/mdx` bundle — is **absent from the after build**.

## One thing this does not fix

A second chunk still matches the markdown-parser fingerprint after the alias:

| Chunk | Before | After |
|---|---:|---:|
| `dd3583a7-*.js` | 1,142,097 B | **gone** |
| `41958015.*.js` | 1,198,209 B | 1,198,207 B |

The alias barely touches `41958015`, so that copy arrives by a different route — most likely the chain in #4892, where 12 files value-import `tinacms` for admin form helpers and pull in `tinacms/dist/index.js`. Worth confirming directly rather than assuming.

Two consequences:

1. **#4892 is still required.** This PR removes one of two doors; the remaining chunk is comparable in size.
2. **Upstream tinacms#7233 would not have covered this on its own.** It repoints `rich-text` at a `sanitize-url` subpath — the same door this PR closes — and leaves `41958015` untouched. Useful context for that PR, and a reason not to simply wait for it.

isaaclombardssw (approved):
This is really ugly but since the fix is just waiting on Tina to come through... ✅

isaaclombardssw:
Tested this branch locally — works great for `next build`/prod, but it breaks `next dev`: any page rendering `<TinaMarkdown>` 500s.

**Cause:** `TINACMS_MDX_SHIM` is an absolute path. Webpack's `resolve.alias` accepts absolute paths, but Turbopack's `resolveAlias` treats the value as relative — it turns `/Users/.../lib/tinacms-mdx-shim.js` into `./Users/.../lib/tinacms-mdx-shim.js` and can't resolve it:

```
Module not found: Can't resolve './Users/.../lib/tinacms-mdx-shim.js'
GET /events/ai-for-business-leaders 500
```

So the webpack alias (prod, CI build, your bundle measurements) is fine — only `next dev` is affected, which is why CI didn't catch it (CI only runs `next build`, never `next dev`).

**Repro:**
```bash
git checkout perf/tina-mdx-sanitize-url
pnpm dev
curl -s -o /dev/null -w "%{http_code}\n" http://localhost:3000/events/ai-for-business-leaders
# -> 500  (dev log: Can't resolve './Users/.../lib/tinacms-mdx-shim.js')
```

**Fix** — point just the Turbopack alias at a project-relative path; webpack keeps the absolute one:
```js
turbopack: {
  resolveAlias: { "@tinacms/mdx": "./lib/tinacms-mdx-shim.js" },
}
```
Confirmed: with that one change the same page returns 200 and rich text renders. Nothing else needed touching — the shim's `sanitizeUrl` is byte-identical to the installed `@tinacms/mdx@2.1.4` and the `dd3583a7` chunk is gone in prod as you measured.

joshbermanssw:
Reproduced and fixed in `014674cb`.

Confirmed the diagnosis before changing anything — Turbopack does prepend `./` to the value it's given:

```
HTTP=500
Module not found: Can't resolve './Users/josh/.../lib/tinacms-mdx-shim.js'
```

And after pointing Turbopack at a relative path, on the same page:

```
✓ Ready in 1384ms
HTTP=200 bytes=625202
dev log resolution errors: (none)
```

Also verified rich text genuinely renders rather than the route merely returning 200 — the block-2 description paragraph is present in the dev-rendered HTML, and no markdown-parser strings appear, so the shim is being used rather than silently falling back.

Kept the two forms deliberately different — absolute for webpack, relative for Turbopack — with a comment saying why, because unifying them re-breaks one bundler or the other. My original comment claimed absolute would "resolve identically" in both; that was the wrong assumption, and it introduced exactly the dev/prod divergence the change was supposed to prevent.

Your point about CI is the more important one: `pr-build.yml` only runs `next build`, so no amount of CI would have caught a Turbopack-only failure. Worth a smoke test that boots `next dev` and asserts a page renders — I can raise that separately rather than expand this PR.

One correction for the record: the installed version here is `@tinacms/mdx@2.1.9`, not 2.1.4 — worth being precise about since the shim header tells the next person to re-diff `sanitizeUrl` against the installed build when the Tina packages are bumped.

PR #4916, opened 2026-07-23
💸 Perf - Mount the Eventbrite Popup only when it opens

### LINKED ISSUE #4904: 💸 Perf - Mount the Eventbrite Popup only when it opens
Sub-issue of the AI for Business Leaders performance PBI. Fully isolated.

**Description**

Blocks 5 and 37 of `content/eventsv2/ai-for-business-leaders.json` are the same "Pick your city" card carousel, four cards each — so eight `EmbeddedCardButton`s render, each with an `eventbriteEventId`.

`components/eventbrite/eventbriteModalButton.tsx` correctly defers the Eventbrite widget script until the user opens a modal — that part is well done and should stay. But the `Popup` itself is mounted unconditionally (the comment notes this is for the close animation), so eight copies of `react-responsive-modal` plus its global stylesheet are live in the tree at all times.

**Proposed solution**

1. Render the `Popup` only once it has been opened — track a "has ever opened" flag so the close animation still plays, but nothing mounts on first paint
2. Alternatively, hoist a single shared modal to the page level and have the buttons drive it, so eight CTAs cost one modal
3. Check whether `react-responsive-modal`'s stylesheet can be scoped rather than imported globally

**Acceptance criteria**

1. No modal DOM or `react-responsive-modal` code is mounted on initial page load
2. Clicking any "Reserve my seat" CTA still opens the Eventbrite checkout, and the open/close animation is unchanged
3. Opening a second card's checkout after closing the first still works (no stacked iframes)
4. The existing script-timeout fallback link still appears when the Eventbrite script fails

isaaclombardssw:
Closes #4904

Each Eventbrite CTA mounted a `react-responsive-modal` `Popup` unconditionally — the old comment kept it always-mounted so the close animation could play. On the AI for Business Leaders page (blocks 5 and 37, eight `EmbeddedCardButton`s) that means eight live modal instances at first paint.

## Change

`components/eventbrite/eventbriteModalButton.tsx` now gates the `Popup` behind a `hasOpened` flag:

- On initial paint `hasOpened` is `false`, so `{hasOpened && <Popup />}` renders nothing — no `react-responsive-modal` instance, no portal, no iframe.
- Clicking a CTA sets `hasOpened` (and `open`) `true`, mounting the Popup and playing the enter animation.
- Once opened it stays mounted, so the close transition still runs; the library unmounts the Popup's children while closed, which already resets the widget.

The deferred script load, the "don't stack a second iframe" guard, the script-timeout fallback link, and the `eventbrite_order_complete` conversion tracking (`sendGTMEvent`) are all untouched.

This is option 1 from the issue (a "has ever opened" flag), the smallest of the three proposed approaches — no page-level refactor, no dep changes.

## Verified

- `pnpm lint` clean (only the pre-existing `z-[2]` Tailwind warning).
- `npx prettier --check` on the changed file clean.
- `npx tsc --noEmit` reports no errors for this file (remaining project errors are the pre-existing `@/tina/client` re-export that only the full dev server emits).
- Code trace: with `hasOpened` false the Popup subtree is not rendered, so no modal DOM or iframe exists until a CTA is clicked; reopening after close re-runs the same open path.

## Left for a human

- A quick manual smoke test in a running dev server (open a CTA, complete/close, open a second card) to confirm the animation feel — the full `pnpm dev` build is memory-heavy in this environment so I verified by code trace and static checks rather than a live click-through.

### DIFF
--- a/components/eventbrite/eventbriteModalButton.tsx
+++ b/components/eventbrite/eventbriteModalButton.tsx
@@ -58,6 +58,8 @@ export const EventbriteModalButton = ({
   const reactId = useId().replace(/[^a-zA-Z0-9_-]/g, "");
   const containerId = `eventbrite-checkout-${eventId}-${reactId}`;
   const [open, setOpen] = useState(false);
+  // Gate the Popup on first open so a closed CTA mounts no react-responsive-modal; once opened it stays mounted for the close animation.
+  const [hasOpened, setHasOpened] = useState(false);
   const [scriptLoaded, setScriptLoaded] = useState(false);
   const [scriptFailed, setScriptFailed] = useState(false);
   // The popup portals its content, so bind via the callback ref once the node mounts, not when `open` flips.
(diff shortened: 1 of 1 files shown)


isaaclombardssw:
This is going to be marginal in terms of perf gains, but harmless cleanup

PR #4931, opened 2026-07-24
🧹 Keep the TinaCMS editor out of the public JS bundle

### LINKED ISSUE #4894: 💸 Perf - AI for Business Leaders event page - LCP 9.3s, TBT 1,250ms on mobile

joshbermanssw:
## TL;DR

Every public page that renders a button was preloading the ~1.5 MB TinaCMS **editor** (`@udecode/plate`, `@tanstack/table`, `mermaid`, `graphql`) — ~494 KiB transferred, 81% unused. Moving one string constant out of a schema file severs the leak, taking `/events/ai-for-business-leaders` from Lighthouse **83 → 91** with **TBT 340ms → 20ms**.

## Why

`templateButton.tsx` (a rendered client component) imported `PROJECT_FORM_MODAL` from `templateButton.schema`. The schema imports the Tina field-editor UI (`ColorPickerInput`/`IconPickerInput`), which pulls the full `tinacms` barrel. So the render chain `imageTextBlock → buttonRow → templateButton → templateButton.schema` dragged the entire editor onto every public page.

Same-host Lighthouse A/B (mobile, headless, identical conditions):

| | Baseline | This PR |
|---|---|---|
| Performance | 83 | **91** |
| Total Blocking Time | 340 ms | **20 ms** |
| Unused JS | 583 KiB | **40 KiB** |
| Editor chunk preloaded | yes | **no** |

Rendered HTML is byte-identical (visible text unchanged, all buttons intact).

## How

`PROJECT_FORM_MODAL` (a plain `"projectFormModal"` string) moves into a new `tinacms`-free `templateButton.constants.ts`. The render path imports from there; the schema re-exports it so `bookingForms` is unchanged. The CMS still gets the full field UI through the schema registry (`tina/config.tsx`), so live editing is untouched.

## Reviewer notes

- **Editor unaffected.** `useTina`/`tinaField` (6 KB, `tinacms/dist/react`) were never the issue; only the field-UI barrel was.
- **~14 other `.schema` files** import the same field UI — this fixes the one on the render path; the rest are a follow-up.

## Tests

Verify `/admin` still edits and saves an event, and button forms still open. Bundle proof: the events page chunk no longer references the editor chunk; `BUNDLE_ANALYSE=true` treemap confirms it's gone.

## Links

Closes #4894

### DIFF
--- a/components/button/templateButton.schema.tsx
+++ b/components/button/templateButton.schema.tsx
@@ -1,10 +1,9 @@
 import { buttonOptions } from "../blocksSubtemplates/tinaFormElements/colourOptions/buttonOptions";
 import { ColorPickerInput } from "../blocksSubtemplates/tinaFormElements/colourSelector";
 import { IconPickerInput } from "../blocksSubtemplates/tinaFormElements/iconSelector";
+import { PROJECT_FORM_MODAL } from "./templateButton.constants";
 
-// Special value: opens the native multi-step "Tell us about your project"
-// modal instead of a JotForm popup. Must not resolve to a JotForm id.
-export const PROJECT_FORM_MODAL = "projectFormModal";
+export { PROJECT_FORM_MODAL };
 
 export const bookingForms = [
   {
--- a/components/button/templateButton.schema.tsx
+++ b/components/button/templateButton.schema.tsx
@@ -3,8 +3,6 @@ import { ColorPickerInput } from "../blocksSubtemplates/tinaFormElements/colourS
 import { IconPickerInput } from "../blocksSubtemplates/tinaFormElements/iconSelector";
 import { PROJECT_FORM_MODAL } from "./templateButton.constants";
 
-export { PROJECT_FORM_MODAL };
-
 export const bookingForms = [
   {
     label: "Tell us about your project (multi-step)",
--- a/components/button/templateButton.tsx
+++ b/components/button/templateButton.tsx
@@ -10,7 +10,7 @@ import Popup from "../popup/popup";
 import { SESSION_STORAGE_KEYS } from "../util/constants";
 import { buttonColorVariants } from "../blocksSubtemplates/tinaFormElements/colourOptions/buttonOptions";
 import RippleButton, { ButtonTinaFields } from "./rippleButtonV2";
-import { PROJECT_FORM_MODAL } from "./templateButton.schema";
+import { PROJECT_FORM_MODAL } from "./templateButton.constants";
 
 type TemplateButtonOptions = {
   buttonText?: string;
--- a/components/button/templateButton.constants.ts
+++ b/components/button/templateButton.constants.ts
@@ -1,0 +1,8 @@
+// Special value: opens the native multi-step "Tell us about your project"
+// modal instead of a JotForm popup. Must not resolve to a JotForm id.
+//
+// Lives in its own tinacms-free module so the render path (templateButton.tsx)
+// can read it without pulling templateButton.schema — which imports the Tina
+// field-editor UI (ColorPickerInput/IconPickerInput) and, through it, the
+// ~1.5 MB tinacms editor barrel. See SSW.Website perf work on the events pages.
+export const PROJECT_FORM_MODAL = "projectFormModal";


components/button/templateButton.schema.tsx
```diff
@@ -1,10 +1,9 @@
 import { buttonOptions } from "../blocksSubtemplates/tinaFormElements/colourOptions/buttonOptions";
 import { ColorPickerInput } from "../blocksSubtemplates/tinaFormElements/colourSelector";
 import { IconPickerInput } from "../blocksSubtemplates/tinaFormElements/iconSelector";
+import { PROJECT_FORM_MODAL } from "./templateButton.constants";
 
-// Special value: opens the native multi-step "Tell us about your project"
-// modal instead of a JotForm popup. Must not resolve to a JotForm id.
-export const PROJECT_FORM_MODAL = "projectFormModal";
+export { PROJECT_FORM_MODAL };
```
tomek-i: why does it need to be exported from here? 
what ever file needs access to it, can it not just import from "templateButton.constants.ts" directly?

components/button/templateButton.tsx
```diff
@@ -10,7 +10,7 @@ import Popup from "../popup/popup";
 import { SESSION_STORAGE_KEYS } from "../util/constants";
 import { buttonColorVariants } from "../blocksSubtemplates/tinaFormElements/colourOptions/buttonOptions";
 import RippleButton, { ButtonTinaFields } from "./rippleButtonV2";
-import { PROJECT_FORM_MODAL } from "./templateButton.schema";
+import { PROJECT_FORM_MODAL } from "./templateButton.constants";
```
tomek-i, line 13: basically that answers my above question, you import from constants so you should remove the re-export from the schema unless there's reasons i am not aware of why it needs to be re-exported from there too

joshbermanssw: Good catch — you're right, nothing imports it from here. Dropped the re-export in 2fe6d1f9; kept the import since `bookingForms` still uses the value.

tomek-i (approved):
LGTM

PR #4936, opened 2026-07-27
⚡ Load Application Insights on demand instead of in every page bundle

### LINKED ISSUE #4903: 💸 Perf - Defer Application Insights init out of the critical window

**Description**

`context/app-insight-client.tsx` constructs the full `@microsoft/applicationinsights-web` SDK in a mount effect on every page, with `enableRequestHeaderTracking`, `enableResponseHeaderTracking`, `autoTrackPageVisitTime` and `autoExceptionInstrumented` all enabled. That work lands squarely in the hydration window, contributing to TBT 1,250 ms and the 17 long tasks.

The SDK also ships as an ES5 build, so it is a likely contributor to Lighthouse's "Legacy JavaScript — Est savings of 41 KiB" (the site's own browserslist is already modern: `chrome >= 111`).

**Proposed solution**

1. Defer `loadAppInsights()` until after the page is interactive — `requestIdleCallback` with a timeout, or a `load` listener
2. Dynamically `import()` the SDK at that point so it isn't in the initial chunk graph
3. Buffer any web-vitals metrics reported by `app/components/web-vitals.tsx` before the SDK is ready, and flush them on init, so no telemetry is lost
4. Review whether request/response header tracking is worth its cost on a static marketing site

**Acceptance criteria**

1. No Application Insights code executes before the page is interactive
2. The SDK is not part of the route's initial JS entry
3. Page views, exceptions and web vitals (LCP, CLS, INP, TTFB, FCP) still arrive in App Insights, including those measured before init
4. TBT measurably improves

joshbermanssw:
## TL;DR

`@microsoft/applicationinsights-web` was imported at module scope by two client components in the root layout, so every page on the site downloaded and parsed ~138 KB of telemetry SDK before hydration. Neither renders anything from it — both construct the SDK inside `useEffect` — so both now import it dynamically.

**Eager JS on `/events/ai-for-business-leaders`: 1,202,372 → 1,063,976 bytes**, one fewer eager chunk, and the SDK moves to an async chunk that no longer blocks hydration.

## Why

That page is 85% JavaScript by transfer size — 1,327 KiB across 77 script requests, against 42 KiB of images and 4 KiB of CSS. Mobile LCP is 89% "render delay" with zero resource load time, i.e. the paint is waiting on the main thread rather than on any download. Telemetry that only runs after mount has no business being in that critical path.

Screenshot 2026-07-27 at 5 30 17 pm

**Figure: Tested lighthouse speed**

## Links

- Closes #4903 — Defer Application Insights init out of the critical window
- Parent: #4894 — Perf: AI for Business Leaders event page

### DIFF
--- a/app/global-error.tsx
+++ b/app/global-error.tsx
@@ -1,8 +1,6 @@
 "use client";
 import { MegaMenuWrapper } from "@/components/server/MegaMenuWrapper";
 import { ErrorPage } from "@/components/util/error-page";
-import { ReactPlugin } from "@microsoft/applicationinsights-react-js";
-import { ApplicationInsights } from "@microsoft/applicationinsights-web";
 import { inter } from "@/lib/fonts";
 import { useEffect } from "react";
 import "styles.css";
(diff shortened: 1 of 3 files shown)


joshbermanssw:
> The only thing that is a bit sus is defining the application insights twice ... can you export it in a hook or does it give problems in the global-error.tsx file? or is it deliberately for slight different config?

Half deliberate, half not — good catch.

I diffed the two configs and the **only** difference is `samplingPercentage`. The provider samples at 20% for cost; `global-error` doesn't, and that one should stay: sampling away four in five fatal errors defeats the point of a crash reporter. The other 13 lines were identical copy-paste.

**A hook won't work**, and it's structural rather than laziness. `global-error.tsx` renders its own `<html>` and `<body>` because Next replaces the entire root layout when that boundary activates — so `AppInsightsProvider` is not in the tree and there is no context to read. It has to construct its own `ReactPlugin`.

A shared factory does the job though, so `lib/app-insights.ts` now exports:

```ts
loadAppInsights({ reactPlugin, samplingPercentage })
```

It owns the config, the dynamic import and the connection-string check, and resolves to `null` when unconfigured so each caller keeps its own log message. Sampling is an explicit optional parameter documented with *why* `global-error` omits it, rather than an invisible difference between two near-identical blobs — which is how it would have drifted next time someone touched one of them.

It also means the dynamic `import()` lives in one place instead of being duplicated across both files, which was a fair criticism of the first commit.

Pushed as `06e70fa7`. `tsc --noEmit` clean, build green, and verified on the preview deploy: eager JS on `/events/ai-for-business-leaders` drops from **1,202,372 to 1,081,434 bytes (−121 KB)**, with the SDK present in none of the 39 eager chunks.

tomek-i (approved):

PR #4939, opened 2026-07-29
💄 Redesign /consulting index + card colour tokens

### LINKED ISSUE #4971: 💄 Redesign the /consulting services index (dark mode, filtering, shared card tokens)
### Problem

The `/consulting` services index is still on the pre-v3 design. It renders as a flat category grid with a `?tag=` query-param filter, doesn't support the light/dark switcher the v3 homepage has, and isn't optimised for mobile.

Specific issues:
- No dark mode. `/consulting` is a major entry point but sits outside the theming that `/` already uses, so the switcher disappears when you land on it.
- The `?tag=` filter triggers a full navigation per click and doesn't scroll to the section.
- Card styling is bespoke rather than token-driven, so it drifts from the homepage service cards.
- No way for content editors to highlight a service as popular.
- Mobile layout is a single narrow column with the filter list stacked above it.

### Suggested change

Redesign the index as a sticky sidebar plus tag-filtered sections:

- Sidebar of categories that filters the grid in place, deep-linkable by hash anchor (`/consulting#consulting-cloud-and-infrastructure`) so the mega menu can point at a category.
- Keep honouring existing `?tag=` URLs, which are indexed and bookmarked.
- Shared `--card` / `--card-hover` design tokens in `styles.css`, exposed to Tailwind, so card surfaces are token-driven in both themes.
- Add `/consulting` to the themed routes so it picks up the light/dark switcher.
- New `popular` boolean on consulting index pages, rendering a POPULAR badge.
- Mobile: sidebar collapses to a sticky filter chip row.

### Notes

Default view must keep rendering every section server-side so the page stays crawlable.

Migrating the homepage `imageCards` block onto the same tokens was tried and backed out — its `opacity`-based hover works for the gradient background presets, which a background-colour token can't, and merging a token after the Tina-configured class silently disables the "Card Background Colour" picker. Worth a separate issue if we still want it.

hveraus:
Redesigns the /consulting services index and introduces shared card colour tokens.

- Affected routes: `/consulting`

- Fixed #4971

**Changes:**
- New card tokens in `styles.css`, exposed to Tailwind as `bg-card` / `bg-card-hover`. Defined in both light (`#f9f9f9` / `#ffffff`) and dark (`#101010` / `#151515`) scopes. Card borders use the existing `--hairline`.
- `/consulting` index redesigned: sticky sidebar, tag-filtered sections, new `ConsultingCard` component. Replaces the old category grid and the `?tag=` query-param filter with hash anchors (old `?tag=` URLs are still honoured).
- `/consulting` opts into the existing home theming system, so it supports the light/dark switcher. Route list is now a single `THEMED_ROUTES` constant shared by the theme boundary, the pre-paint script and the mega menu.
- Mobile layout: sidebar collapses to a sticky chip row.
- New `popular` boolean on consulting index pages, rendering a POPULAR badge.
- Section-building logic lives in `helpers/consultingSections.ts` and is unit-tested.
- Deletes `components/consulting/index/{category,pageCard,tag}.tsx` and the framer-motion card animation they carried.

**Not included:** the homepage `imageCards` block was briefly moved onto the new tokens and has been reverted. Its `dark:opacity-60 → group-hover:opacity-100` mechanism works for gradient presets, which a background-colour token cannot, and merging a token after the Tina-configured class silently disabled the "Card Background Colour" picker. Worth a focused follow-up rather than riding along here.


app/consulting/index.module.css
joshbermanssw, line 1: We stick away form using these .index.module.css files 

Also creates a BUNCH of new vars that may just be used in this directory.

If we create css variables they should be global

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
 import { tinaField } from "tinacms/dist/react";
+import styles from "./index.module.css";
 
 const allServices = "All SSW Services";
 
-export default function ConsultingIndex({ tinaProps }) {
-  const gridRef = useRef(null);
-  const router = useRouter();
-  const [selectedTag, setSelectedTag] = useState(allServices);
+const slugify = (value: string) =>
```
joshbermanssw: do we have a utility function for this all ready?

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+  value
+    .toLowerCase()
+    .replace(/[^a-z0-9\s-]/g, "")
+    .trim()
+    .replace(/\s+/g, "-");
 
-  const filterDidChange = useRef<boolean>(false);
+const mapPageUrl = (page) =>
```
joshbermanssw: do we have a utility function for this all ready?

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
-  const categories = useMemo(() => {
-    return node.categories.reduce((acc, curr) => {
-      const mappedPages = curr.pages.reduce((pageAcc, p) => {
-        // Skip entries with neither an external URL nor a linked page —
-        // otherwise reading p.page.id below crashes the build prerender.
-        if (!p.externalUrl && !p.page?.id) {
-          return pageAcc;
+  const [activeSectionId, setActiveSectionId] = useState<string>("");
+
+  const tags = useMemo(() => {
```
joshbermanssw: (this wasn't you but https://www.ssw.com.au/rules/follow-boy-scout-rule) 

all 4 memos can be combined into one useMemo hook instead of a chain of 4 of them

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+                    node.sidebar[section.index],
+                    "label"
+                  )}
+                >
+                  {section.label}
+                </h2>
+
+                <div className={styles.cardGrid}>
+                  {section.pages.map((page) => (
+                    <a
```
joshbermanssw: I would move this <a> and all of it's descendents into their own component that can be re-used... i.e. consulting card or something along those lines

joshbermanssw (commented):
couple changes

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+      );
+      return {
+        ...tag,
+        pages,
+      };
+    });
+  }, [allMappedPages, tags]);
+
+  const contentSections = useMemo(
+    () => sections.filter((section) => section.name !== allServices),
```
joshbermanssw: **6 services disappear from the page here.**

Sections come from `node.sidebar` and this drops the `All SSW Services` entry, so a card only renders if it carries a real tag. But 6 pages in `content/consulting/index/index.json` have no `tags` key at all, so they hit the `: [allServices]` fallback on line 64 and now match no section:

- Azure DevOps to GitHub Migration
- Odoo
- Clean Architecture
- Vertical Slice Architecture
- ServiceNow
- Chinafy - Mini Apps

They render today because `selectedTag` defaulted to `allServices`, so the old page showed them under their category heading.

Either tag those 6 in the content, or keep a fallback section for anything that matched nothing. Making `tags` required in the Tina schema would stop this recurring silently the next time someone adds a card.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
-    () =>
-      node.sidebar?.map((item) => {
-        return {
-          label: item.label,
-          name: item.tag?.name,
-        };
-      }),
-    [node]
+    });
+    return Array.from(unique.values());
```
joshbermanssw: **This dedup drops distinct cards, not duplicates.**

Keying on `url` alone assumes one card per page reference, but two entries in the content share `content/consulting/azure.mdx` with different titles, descriptions, logos and tags:

- "Azure AI" (tags: `ai-development`)
- "Microsoft Azure" (tags: `cloud-and-infrastructure`, `platform-development`)

First-wins keeps Azure AI, so **Microsoft Azure vanishes from both the Cloud and Infrastructure and Platform Development sections**. Same shape for `content/consulting/sharepoint.mdx`: "SharePoint" wins and **"Document Management" disappears from Other SSW Services**.

Keying on the rendered card instead of the page (`${page.url}|${page.title}`) fixes it. Worth checking whether cross-category dedup is needed at all, since sections are tag-driven now rather than category-driven.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+        document.getElementById(fromHash.sectionId)?.scrollIntoView({
+          behavior: "smooth",
+          block: "start",
+        });
+      });
+    } else {
+      setActiveSectionId(contentSections[0].sectionId);
+    }
 
-    const tagParam = searchParams.get("tag");
```
joshbermanssw, line 80: Removing the `?tag=` reader breaks two live links in the primary nav. `content/megamenu/menu.json` still points at:

- line 38: `/consulting?tag=Platform-Development`
- line 65: `/consulting?tag=Cloud-and+Infrastructure`

Both now land at the top of the page with no filter and no scroll. They want updating to the new anchors (`/consulting#consulting-platform-development` and `/consulting#consulting-cloud-and-infrastructure`).

These URLs are public and indexed, so a small `?tag=` to hash shim on mount would also be worth it for anyone arriving from an external link or a bookmark.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+      block: "start",
+    });
+  };
 
   return (
-    <>
-      <Container className="flex-1 pt-2">
-        <div className="min-h-8 w-full max-w-9xl md:min-h-12">
+    <div
+      className={cn(styles.shell, isDark && "dark")}
```
joshbermanssw: This shell gets `.dark` from React state only, and nothing renders `HomeThemePrePaint` inside it. `HomeThemePrePaint` only appears in `HomeThemeShell` (homepage) and in `MegaMenuWrapper`, so the menu is protected but this page is not: the SSR'd HTML ships with `DEFAULT_THEME = "dark"` and light-mode users will see the page body flip on hydration.

Wrapping the content in `HomeThemeShell` instead of hand-rolling the shell fixes the flash, and gets you `bg-background text-foreground` plus one theming mechanism rather than two (`.dark` and `data-theme` are both being set here).

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+      data-theme={isDark ? "dark" : "light"}
+    >
+      <main className={styles.page}>
```
joshbermanssw: Nested `<main>`. `components/layout/layout.tsx:130` already wraps every page in `<main className="grow bg-white">`, so this is a second `main` landmark inside the first, which isn't valid HTML. A `div` does the job here.

While you're in there: that outer wrapper's `bg-white` is unconditional, and `.shell`'s `min-height: 100%` resolves against a flex `grow` parent, which is fragile. Worth a look in dark mode on a tall screen to confirm no white band shows below the last section.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+    } else {
+      setActiveSectionId(contentSections[0].sectionId);
+    }
 
-    const tagParam = searchParams.get("tag");
-    const query = getSelectedTagFromQuery(tagParam);
-    setSelectedTag(query || allServices);
-  }, [searchParams]);
+    return () => observer.disconnect();
+  }, [contentSections]);
```
joshbermanssw, line 95: `contentSections` memoises on `node`, and `useTina` hands back a fresh object on every edit, so in the visual editor this effect re-runs on each keystroke. That resets `activeSectionId` to the first section and, when the URL carries a hash, re-fires `scrollIntoView` and yanks the editor's viewport back.

Memoising on something stable (the sidebar/categories content rather than the node identity) would settle it.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+
+    const hash = window.location.hash.replace("#", "");
+    const fromHash = contentSections.find(
+      (section) => section.sectionId === hash
+    );
+    if (fromHash) {
+      setActiveSectionId(fromHash.sectionId);
+      window.requestAnimationFrame(() => {
+        document.getElementById(fromHash.sectionId)?.scrollIntoView({
+          behavior: "smooth",
```
joshbermanssw: `behavior: "smooth"` here and on line 148 ignores `prefers-reduced-motion`. The CSS module has a reduced-motion block, but it only covers transitions, not programmatic scrolling.

A `window.matchMedia("(prefers-reduced-motion: reduce)").matches` check picking `"auto"` over `"smooth"` covers both call sites.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
-          <div>
-            <div
-              ref={gridRef}
-              className="mt-4 grid grid-cols-1 gap-2 lg:grid-cols-2"
-            >
-              <FilterContextProvider filterDidChange={filterDidChange.current}>
-                {categories.map((category, index) => {
+        <div className={styles.layout}>
+          <aside className={styles.sidebar}>
+            <h1 className={styles.heading}>Consulting Services</h1>
```
joshbermanssw: The page's `<h1>` is inside `<aside>`, so the document title sits in the complementary landmark and `<main>` has no heading at all. Screen reader users navigating by landmark will find "Consulting Services" under "complementary".

Moving the `<h1>` out to sit above the layout grid (or making this wrapper a plain `div`) sorts it.

app/consulting/index.tsx
```diff
@@ -1,186 +1,274 @@
...
+                        node.sidebar[section.index],
+                        "label"
+                      )}
+                    >
+                      <a
+                        href={`#${section.sectionId}`}
+                        onClick={(event) =>
+                          onNavClick(event, section.sectionId)
+                        }
+                        aria-current={isActive ? "true" : undefined}
```
joshbermanssw: For in-page anchor nav this should be `aria-current="location"` rather than `"true"`. `"true"` is valid ARIA but it reads as "current item" generically; `"location"` is the value specified for "the current position within a flow", which is what the scroll-spy is expressing.

components/layout/homeTheme.tsx
```diff
@@ -19,6 +19,9 @@ const STORAGE_KEY = "ssw-home-theme";
 // choice or `prefers-color-scheme`.
 const DEFAULT_THEME: HomeThemeMode = "dark";
 
+const isThemedRoute = (pathname: string) =>
```
joshbermanssw: The themed-route list now lives in three places that have to stay in sync: this helper, the `PRE_PAINT_SCRIPT` string on line 71, and a hand-rolled copy of the same OR in `MegaMenuWrapper.tsx:21`.

Exporting a single `THEMED_ROUTES` array, deriving both this helper and the script's inlined check from it, and importing `isThemedRoute` in `MegaMenuWrapper` would leave one place to edit when the next route gets themed.

components/server/MegaMenuWrapper.tsx
```diff
@@ -18,7 +18,7 @@ export function MegaMenuWrapper(props) {
   // into theming: the dark toggle and the `.dark` scope below, whose token
   // utilities (bg-background/text-foreground/border-hairline) flip via the CSS
   // variables under the `dark` class on this element.
-  const isHome = pathName === "/";
+  const isHome = pathName === "/" || pathName === "/consulting";
```
joshbermanssw: `isHome` is no longer accurate now that it covers `/consulting`, and it's a duplicate of `isThemedRoute` in `homeTheme.tsx:22` (not currently exported). Importing that helper and renaming this to `isThemedRoute` would keep the two lists from drifting apart, since this one also gates `HomeThemePrePaint` on line 38.

styles.css
```diff
@@ -185,6 +185,8 @@
     --background-sunken: black;
     --background-raised: #262626;
     --background-overlay: #44403c;
+    --card: #101010;
+    --card-hover: #151515;
```
joshbermanssw, line 196: These two are defined only in the dark block, with no light counterpart, unlike every token around them (`--hairline` is set in both, at line 108 and line 177). So `bg-card` resolves to nothing in light mode.

The only consumer today guards it (`bg-white dark:bg-card`), so nothing is visibly broken, but for a PR whose headline is shared card tokens it's worth adding the light values now. Otherwise the next person to reach for `bg-card` gets a transparent card and no clue why.

joshbermanssw (commented):
Nice direction on the redesign, and the `imageCards` change is a clean simplification (opacity fade to a real colour transition on shared tokens).

Two things I'd want fixed before this merges, both content losses rather than styling:

1. **8 cards stop rendering.** 6 pages carry no `tags` at all, so they fall to the `[allServices]` default and now match no section. On top of that, the new URL dedup drops two distinct cards (Microsoft Azure, Document Management) because they share a page reference with a differently-titled card. Details inline.
2. **The mega menu's two `/consulting?tag=...` deep links are dead**, since the query-param reader is gone. `content/megamenu/menu.json` lines 38 and 65 need rewriting to the new hashes.

Rest of the comments are smaller: the themed-route list is now duplicated in three places, `/consulting` has no pre-paint script so it will flash dark to light for light-mode users, and there's a nested `<main>`.

I've deliberately left the `index.module.css` file alone here given my earlier comment on it. Worth knowing for whatever replaces it: the inactive nav colour (`#aaaaaa` on `#fafafa`) is 2.2:1 in light mode against a 4.5:1 requirement, and the POPULAR badge lands around 3.2:1 in dark, so both need new values wherever those styles end up.

app/consulting/index.module.css
hveraus: fixed, not using .index.module.css files any more

app/consulting/index.tsx
hveraus: Claude did it, need a test please

app/consulting/index.tsx
hveraus: fixed, looks good now.

app/consulting/index.tsx
hveraus: fixed, looks good now.

styles.css
hveraus: fixed with Claude, commit https://github.com/SSWConsulting/SSW.Website/pull/4939/changes/36ae5e583ebe9e82dc9a96a2660d7518d1036823

app/consulting/index.tsx
hveraus: fixed with Claude - https://github.com/SSWConsulting/SSW.Website/pull/4939/changes/016e5086f0eb3572b9d3dc7bc2db92dd1cae58e6

app/consulting/index.tsx
hveraus: Fixed with Claude - https://github.com/SSWConsulting/SSW.Website/pull/4939/changes/9f77db86e8225ae52ef346ab40ac48e19d008c6b

tiagov8:
Hi @hveraus and @joshbermanssw 

I know this is more about design, but...

1. Should we create a new category to reduce the many "Other Services" list?

E.g.

**Software Engineering & Delivery**

- Azure DevOps to GitHub Migration
- Clean Architecture
- Vertical Slice Architecture
- Strategic Architecture
- Software Auditing
- ALM + GitHub + Azure DevOps
- DevOps
- Scrum
- Kubernetes
- Access Database Upsizing
- Test Practice

--

Screenshot 2026-08-05 at 2 56 58 PM

**Figure: Way too many "Other Services"**

tiagov8:
@hveraus 

I recorded [this video](https://sswcom-my.sharepoint.com/:v:/g/personal/tiagoaraujo_ssw_com_au/IQB__pLNT18OTYqYUO_VgQGPAcNpakhZ74CBsZO-KEDPVKY?e=ZXPVTV&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D) with some thoughts.

Here is an AI summary:

## Summary

The overall feedback was positive. The main concerns were consistency between light and dark mode, text contrast, repeated labels, and outdated services being promoted.

## Action points

1. **Make card hover states consistent**

   * Remove the background-color change used in light mode.
   * Keep the red arrow animation.
   * Use a border-color change for hover in both light and dark mode.

2. **Improve text contrast**

   * Darken inactive sidebar items in light mode.
   * Lighten breadcrumbs and other secondary text in dark mode.
   * Use consistent, theme-specific gray values.

3. **Remove repeated “SSW” wording**

   * Change `All SSW Services` to `All Services`.
   * Change `Other SSW Services` to `Other Services`.
   * Rename the heading to `Other Services`.

4. **Remove or temporarily hide SSW Digital**

   * Its website looks outdated and should not be promoted until it has been refreshed.

5. **Check category navigation**

   * Keep the distinction clear: `All Services` displays everything, while an individual category displays only services within that category.
   * Ensure the active category remains visually clear, including on hover.


--

PS. Not a normal YakShave as we don't want a separate PBI for this

tiagov8 (changes requested):
- [x] 1. Please action this comments: https://github.com/SSWConsulting/SSW.Website/pull/4939#issuecomment-5198014001 
- [x] 2. And this https://github.com/SSWConsulting/SSW.Website/pull/4939#issuecomment-5197860657

app/consulting/index.tsx
isaaclombardssw: Ken please double check this :)

app/consulting/index.tsx
isaaclombardssw: ditto

hveraus:
> Hi @hveraus and @joshbermanssw
> 
> I know this is more about design, but...
> 
> 1. Should we create a new category to reduce the many "Other Services" list?
> 
> E.g.
> 
> **Software Engineering & Delivery**
> 
> * Azure DevOps to GitHub Migration
> * Clean Architecture
> * Vertical Slice Architecture
> * Strategic Architecture
> * Software Auditing
> * ALM + GitHub + Azure DevOps
> * DevOps
> * Scrum
> * Kubernetes
> * Access Database Upsizing
> * Test Practice
> 

There are too many items on that indexing page, we should add a tracker to review how many of them are not clicked, and hide them, we should do it in another PBI

hveraus:
> ## Action points
> 1. **Make card hover states consistent**
>    
>    * Remove the background-color change used in light mode.
>    * Keep the red arrow animation.
>    * Use a border-color change for hover in both light and dark mode.

Done. except the first one, there are subtle background change on both dark and light mode, I think it's nice. 

> 2. **Improve text contrast**
>    
>    * Darken inactive sidebar items in light mode.
>    * Lighten breadcrumbs and other secondary text in dark mode.
>    * Use consistent, theme-specific gray values.

Done. 


> 3. **Remove repeated “SSW” wording**
>    
>    * Change `All SSW Services` to `All Services`.
>    * Change `Other SSW Services` to `Other Services`.
>    * Rename the heading to `Other Services`.

Done. 


> 4. **Remove or temporarily hide SSW Digital**
>    
>    * Its website looks outdated and should not be promoted until it has been refreshed.

Not done, we'll clean up the index page in another PBI. 


> 5. **Check category navigation**
>    
>    * Keep the distinction clear: `All Services` displays everything, while an individual category displays only services within that category.
>    * Ensure the active category remains visually clear, including on hover.

Done.

isaaclombardssw (approved):
✅ This is good to me, but would be good to get another review?

babakamyljanovssw (approved):
LGTM

tiagov8 (approved):

PR #4942, opened 2026-07-30
⬆️ Upgrade to Next.js 16

joshbermanssw:
## TL;DR

Upgrades Next.js 15.5.18 → 16.2.12 per the [official v16 guide](https://nextjs.org/docs/app/guides/upgrading/version-16), and clears the two webpack-only dependencies that were the only thing stopping `next build` from running on Turbopack (the new default). Async request APIs were already migrated in the 15 upgrade, so no page/route signatures change.

**Files to review (9, excluding the lockfile):**

| File | Why |
|---|---|
| `next.config.mjs` *(start here)* | The `webpack()` block and `eslint.ignoreDuringBuilds` are gone. |
| `proxy.ts` | Was `middleware.ts`; export renamed `middleware` → `proxy`. |
| `lib/platform.js` *(new)* | Replaces the `next-plugin-preval` module. |
| `tailwind.config.js` | Now imports `platform` instead of declaring it. |
| `components/blocks/colorPalette.tsx` | Reads `platform` directly; builds `bg-platform-*` inline. |
| `package.json` | Version bumps; drops two deps; `--turbopack` removed from `dev`. |

## Why

Next 16 makes Turbopack the default for `next build`, and **fails the build outright if a `webpack` config is present**. Ours had exactly two things in it, and neither needed to be:

- The `@svgr/webpack` rule had no callers — every SVG in this repo is referenced by public path (`/images/...`), not imported as a module, so no `.svg` ever hit that loader.
- The `@tinacms/mdx` shim alias was already mirrored in `turbopack.resolveAlias`, because dev has run on Turbopack for a while.

`next-plugin-preval` is a webpack loader, so it was already inert under `next dev`, and its peer range caps out at Next 15. All it did was build-time-evaluate a nine-line `.map()` over the Tailwind platform list. That list now lives in `lib/platform.js` — one source of truth, imported by `tailwind.config.js` and the component, with no bundler involvement.

## Reviewer notes

- **`proxy` runs on Node, not edge.** Next 16 does not support the edge runtime in `proxy`. This one only sets response headers, so there is nothing edge-specific to lose.
- **`middleware/noIndex.ts` → `lib/noIndex.ts`.** Moved so no `middleware/` directory sits beside `proxy.ts` and risks resolving as the convention.
- **`bg-platform-*` is still safelisted.** `tailwind.config.js` safelists it from the same array, so the dynamic class name resolves exactly as the preval'd `className` did.
- **`eslint-config-next` deliberately left at 15.3.5.** Bumping it to 16 pulls in `eslint-plugin-react-hooks` v7 and floods this PR with unrelated lint errors. Nothing in Next 16 requires it — `next lint` was already unused here.
- **Bundle analysis needs a flag now.** `BUNDLE_ANALYSE=true` builds must use `next build --webpack`; the analyzer no-ops under Turbopack and prints that instruction itself.

## Tests

`pnpm exec jest --testPathPatterns '^__tests__/'` — 19/19 pass. `pnpm build` and the UI/Lighthouse suites are the real gate for the Turbopack switch; leaving those to CI.

## Follow-up

- Bump `eslint-config-next` / `@next/eslint-plugin-next` to 16 and fix the `react-hooks` v7 fallout.
- Consider `experimental.turbopackFileSystemCacheForDev` for faster dev restarts.
- `webpack-require-context.d.ts` is now dead (no `require.context` callers); safe to delete separately.

## Links

- [Next.js 16 upgrade guide](https://nextjs.org/docs/app/guides/upgrading/version-16)


bradystroud (approved):
Tested a bunch of different pages and looks good. Page speed inside scores look pretty good as well. Awesome that we could delete lots of code.

Issue #4767, opened 2026-06-16
Staging slots can orphan and exhaust the Container registry limit (prod outage)


isaaclombardssw:
## What happened

Production (`app-sswwebsite-9eb3`) went down on 2026-06-15/16. Old PR staging slots had accumulated and exhausted the container registry storage quota and API rate limits, so production cold-starts failed.

The Azure web app service platform surfaced this as a misleading image-pull error:

```
Failed to pull image: mcr.microsoft.com/appsvc/msitokenservice:stage5.
ImageNotFoundFailure (and earlier: 401/403 failed to fetch token)
```

This sent debugging down the wrong path (MCR outage, networking, managed identity) when the real cause was an authentication bottleneck and registry rate/bandwidth limits with the container registry. Killing the old staging deployments restored the site.

## Why the slots piled up

- `pr-close-delete-env.yml` deletes a PR's slot only as its **last** step, after several ACR / Key Vault / role-assignment steps that can fail first — so any earlier error leaves the slot behind (silent orphan).
- The delete matrix has no `fail-fast: false`, so in the weekly batch one failing slot **cancels cleanup of all the others**.
- Cleanup isn't idempotent: running it on an already-deleted slot errors instead of no-op'ing.
- The only backstop (`weekly-slots-cleanup.yml`) runs **weekly**, so an orphan can linger up to 7 days.
- **On-close cleanup silently skips conflicted PRs (confirmed).** GitHub does not dispatch a `pull_request` workflow run when a PR has no mergeable ref (a merge conflict) — long-stale PRs almost always conflict. So closing a months-old conflicted PR fires the `closed` event but creates **no `pr-close-delete-env` run at all** (no failed run — none), and the slot is orphaned silently. Verified: the PRs whose close did not trigger cleanup were exactly those missing `refs/pull/<n>/merge`; mergeable PRs trigger fine. (Mitigated by the reconcile-based weekly sweep below, which deletes by slot-vs-PR-state and does not depend on the close event.)
- Nothing stops new slots being created once the plan is already full.
- Minor correctness bugs: `az acr show` is passed the login-server FQDN instead of `ACR_NAME`; `az acr repository show-manifests` is deprecated.

## Resolution (this PR)

[...]

PR #4946, opened 2026-07-30
🧹 Stop PR builds pushing throwaway images to ACR

joshbermanssw:
## TL;DR

`pr-build.yml` runs on every `pull_request` and pushed a `pr-<n>` image to ACR, so the registry holds 11 `pr-*` tags right now even though staging slots are capped at 5 and gated behind `/deploy`. PR builds now validate the Docker build without pushing it.

## Why

The `MAX_PR_SLOTS=5` cap added in #4835 gates App Service *slots*, not registry images — nothing was ever capped on the ACR side. Images are only deleted when their PR closes, so every long-lived open PR keeps a few hundred MB of dormant image on the shelf indefinitely.

## Reviewer notes

- **`/deploy` is not slower.** BuildKit's `cache-to` export is independent of `push`, so PR builds keep warming the shared `buildcache` tag.
- **Default is `true`.** `main-build-and-deploy.yml` and `pr-deploy-command.yml` omit the input and still push.
- **Backlog is untouched.** The existing 11 tags belong to open PRs, so the Monday sweep skips them — draining those is a separate change.

## Links

- [#4767 — staging slots are a limited resource](https://github.com/SSWConsulting/SSW.Website/issues/4767)
- [#4835 — /deploy opt-in + slot cap](https://github.com/SSWConsulting/SSW.Website/pull/4835)

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -15,6 +15,8 @@ jobs:
     uses: ./.github/workflows/template-build.yml
     with:
       tag: pr-${{ github.event.number }}
+      # Validate the build only - `/deploy` is what pushes a pr-<n> image to ACR.
+      push: false
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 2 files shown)


18-th (approved):
LGTM

PR #4950, opened 2026-07-30
💄 Migrate /consulting/artificial-intelligence to the v3 block set

joshbermanssw:
## TL;DR

The AI consulting page was still on the v2 blocks (`imageTextBlock`, `spacer`, placeholder logos from `/images/Louis Test Images/`), so it looked nothing like the v3 treatment shipped on `/consulting/react`. This rebuilds it on v3, and adds the three pieces of v3 that were missing to do it properly: a Pills block, an animated hero composition, and a red glow that bleeds past block edges.

**Files to review (14, +615 / -355):**

| File | Why |
|---|---|
| `content/consultingv2/artificial-intelligence.json` *(start here)* | The whole page, rewritten block-for-block. |
| `components/blocks/v3/pills/` *(new)* | New `v3Pills` block — schema + component. |
| `components/blocks/v3/hero/media/AiConsultingHeroMedia.tsx` *(new)* | Animated neural network for the hero RHS. |
| `components/layout/v2ComponentWrapper.tsx` + `tailwind.config.js` | Red glow now bleeds past the block when `bleed` is on. |
| `components/blocks/v3/testimonials/*` | Case-study link text is editable instead of hardcoded. |
| `components/blocks/index.ts`, `blocks-renderer.tsx`, `hero/media/registry.tsx` | Registration for the above. |

## Why

Second page onto v3, after `/consulting/react`. Sequencing follows the React page so the two read as one system: problem → why SSW → process → lead capture → proof → services → FAQ → CTA.

Two pieces of the old page were worth keeping and are carried over into v3 blocks rather than dropped — the FPE case study (now the `v3Testimonials` slide, using the real Thomas Bidou quote already in `content/testimonials/testimonials.json`) and YakShaver (now a second `v3VideoHighlights` block).

## Reviewer notes

- **Copy needs a second opinion.** The FAQ answers on data/training and model choice make claims that should be checked before this goes live.
- **`v3Pills` is deliberately generic.** Brow + heading + a wrapping row of labels, each with an optional image and link. Named for its shape, not for "tech we use", so it can carry industries or certifications later.
- **Glow bleed is opt-in and isolated.** Only fires when `bleed` *and* `redGlow` are both set; the non-bleed path is byte-identical to before.
- **Bleed uses its own gradients.** Growing the layer past the block would otherwise drag the glow's anchor with it, so the bleed variants re-anchor to the block's real corner and shrink their stops to offset the taller gradient box. Those stop percentages are eyeballed, not derived.
- **No new hero block.** `mediaType` resolves through the existing registry, same mechanism as the React atom. Its animation parks under reduced motion.
- **Placeholder logos gone.** Now uses the same `/images/ssw-web/logos/` set as the React page.

## Links

- Reference implementation: [/consulting/react](https://www.ssw.com.au/consulting/react)


isaaclombardssw (approved):
LGTM

PR #4978, opened 2026-08-11
Redesign the event page template

### LINKED ISSUE #4864: ✨ Events UI - Improve event page UX and visual consistency
Requested by: @tiagov8 via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @wicksipedia, @tiagov8, @adamcogan, @designbyalex, @vladislav-kir, @isaaclombardssw, @babakamyljanovssw @Calinator444 @0xharkirat 

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (5 min 18 sec)](https://www.youtube.com/watch?v=hG5KjFZ6GGM)

&nbsp;&nbsp;Url from screen share: [https://claude.ai/design/p/4c9d6630-b08a-4705-af49-c429aab57bb0?file=Event+Page.dc.html](https://claude.ai/design/p/4c9d6630-b08a-4705-af49-c429aab57bb0?file=Event+Page.dc.html)

### Pain
Improve the UI/UX and visual consistency of the event pages on the ⭐ SSW Website (Dev - Next.js). The current implementation has inconsistent styling, unclear CTAs, missing agenda support, inconsistent speaker presentation, markdown rendering issues, and layout problems for multi-day events. A proposed redesign has been provided and should be used as guidance for implementation.

Requested improvements include:
- Replace unclear CTA labels such as "Find out more" with registration-focused wording
- Fix markdown rendering/layout issues caused by spacing
- Standardize heading styles, punctuation, and colors
- Make speaker images clickable and link to speaker profiles
- Improve event hero/header design to better align with SSW branding
- Display speaker names prominently in the hero section
- Improve readability for multi-day events and agenda layouts
- Standardize speaker image sizing and shapes
- Add a clear event summary/sidebar containing key event information
- Add support for agenda display on event pages
- Add share event and add-to-calendar actions
- Add optional "limited spots" messaging when attendee counts are unavailable
- Ensure links are visually identifiable (e.g. underline styling)
- Apply the improved design consistently across all new event pages

The HTML/design reference will be attached to the GitHub issue. Coordinate with Tiago and Alex for design clarification if required.

### Acceptance Criteria
1. Event CTA labels use clear registration-focused wording instead of generic labels such as "Find out more"
2. Markdown content renders correctly without spacing/layout issues
3. Headings across event pages use consistent styling, punctuation, and colors
4. Speaker profile images are clickable and navigate to the corresponding speaker profile page
5. Speaker names are displayed prominently in the event hero/header section
6. Speaker avatars/images are displayed consistently across all event pages
7. Multi-day events display dates and agenda information in a readable and visually consistent format
8. Event pages include an agenda section when agenda data exists
9. Event pages include a sidebar or summary section containing key event information such as dates, location, pricing, and CTA
10. Event pages support share event and add-to-calendar actions
11. Optional urgency messaging such as "Limited spots" can be displayed without requiring attendee count data
12. Links are visually distinguishable from regular text
13. Updated event page styling aligns with SSW branding and approved design references
14. Changes are applied consistently across all new event pages

--

15. **Extra task:** I think the URLs should be more specific by including the month, the city, then the extra

E.g. from

> https://www.ssw.com.au/events/2026/storage-wars
> https://www.ssw.com.au/events/2026/angularhackdaymelb26
To
> https://www.ssw.com.au/events/2026-07/brisbane/user-group-storage-wars
> https://www.ssw.com.au/events/2026-10/melbourne/angular-hackday

Please update the URLs to include the months and more about the event

--

### Screenshot
Screenshot
**Figure: Proposed redesigned event page with speaker names, clearer CTA, event summary, agenda, and sidebar details**

tiagov8:
Closes #4864

Redesigns the calendar event page (`events-preview.tsx`) to the new SSW design.

## Changes
- **Hero** — status/type/entry label, bold title, date · time · location, register CTA, and speaker avatars. Past events show a muted status and a "Watch the recording" CTA.
- **About** — red eyebrow, large lead paragraph, and body prose.
- **Speakers** — card grid with hover (avatar, role, bio, design-system arrow).
- **Sidebar** — sticky event-details card: banner, entry, date chip, venue, register, add-to-calendar (.ics) + share, availability, and "Hosted by SSW".
- Page width capped at `max-w-8xl`.

## New TinaCMS fields (eventsCalendar)
`entryCost`, `venue`, `ctaLabel`, `price`, `lead`, `availability`, `bannerImage`, and a per-presenter `role`. Removed the now-unused `headerLayout` field.

The sample event (Angular Hack Day – Melbourne) is populated to demo the layout.


tiagov8:
📌 Note: the **Agenda** section is not included here. It will be added in a separate, dedicated PR to keep this one focused and simplify implementation.

tiagov8:
FROM

Screenshot 2026-08-11 at 12 27 56 PM

---

TO

Screenshot 2026-08-11 at 5 14 45 PM

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +37,322 @@ type EventsPreviewProps = {
   tinaProps: { data: object };
 };
 
+// Circular speaker portrait with the SSW-red ring used in the hero.
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **The comment says red, the code says grey.**

Line 54 is `ring-gray-200`. Commit 99de7298 changed the ring to a neutral grey and this line stayed behind.

```suggestion
// Circular speaker portrait with a neutral grey ring, used in the hero.
```

0xharkirat (commented):
### Notes, not blockers

- **13 content files still have `headerLayout`.** The schema drops the field but `content/events-calendar/` still carries it in 13 JSON files. The build passes, so nothing is on fire. But Tina will quietly strip the field the next time anyone edits one of those events in the CMS, and that lands as a random unrelated diff in someone else's PR. Cleaner to sweep them here.

- **`dateChip` has the same SSR problem as `isPast`.** It is set inside the same `useEffect` in `useFormatDates`, so the red date chip in the sidebar paints as an empty red square before it fills in. Same root cause as the `isPast` comment, so probably worth doing together.

- **The .ics has no line folding.** RFC 5545 wants lines folded past 75 characters, and a long title plus a long venue will go over. Only some calendar clients care, so low priority.

- **An event that finished this morning still reads as "Today".** `formatRelativeEventDate` returns `TODAY` for same-day events, so `isPast` stays false and the recording link does not show up until tomorrow. Might be exactly what you want, just flagging it.

- **Not yours, but you will trip over it: the event page has an infinite render loop.** I ran this branch locally and the console fills with `Maximum update depth exceeded`, traced to the `useEffect` in `useFormatDates`. The caller builds `new Date(...)` on every render and those objects are the effect's dependencies, so the effect re-runs forever.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +37,322 @@ type EventsPreviewProps = {
...
+}
+
+// Escape reserved characters for an iCalendar text value.
+function escapeIcs(text: string) {
+  return (text || "")
+    .replace(/\\/g, "\\\\")
+    .replace(/;/g, "\\;")
+    .replace(/,/g, "\\,")
+    .replace(/\n/g, "\\n");
+}
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **The calendar download can come out malformed.**

`\r` is not handled here. If any Tina field has Windows line endings, a bare carriage return ends up inside the .ics value, which RFC 5545 does not allow. Some calendar apps will refuse the file rather than fix it.

```suggestion
// Escape reserved characters for an iCalendar text value.
function escapeIcs(text: string) {
  return (text || "")
    .replace(/\\/g, "\\\\")
    .replace(/;/g, "\\;")
    .replace(/,/g, "\\,")
    .replace(/\r\n?/g, "\n")
    .replace(/\n/g, "\\n");
}
```

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +37,322 @@ type EventsPreviewProps = {
...
+  ctaHref,
+  dateChip,
+  dateLine,
+  timeLine,
+}: EventSidebarProps) {
+  const [copied, setCopied] = useState(false);
+
+  const city = event.cityOther || event.city;
+  const state = CITY_MAP[event.city]?.state;
+  const venueUrl = CITY_MAP[event.city]?.url;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **A custom venue links to the wrong place.**

Set `venue` to something that is not an SSW chapel, say "Melbourne Convention Centre", and the sidebar shows that name but the link still goes to SSW Chapel Melbourne. Wrong building.

The name on the next line honours `event.venue`, but the URL here always comes from `CITY_MAP`. This PR adds the `venue` field for exactly this case, so it gets hit as soon as anyone uses it.

```suggestion
  const venueUrl = event.venue ? null : CITY_MAP[event.city]?.url;
```

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +37,322 @@ type EventsPreviewProps = {
...
+  const venueName = event.venue || CITY_MAP[event.city]?.name || city;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **The venue name repeats the city.**

For a city we have no chapel in, with no venue set, the sidebar prints the city name twice, stacked:

```
Gold Coast
Gold Coast
```

`CITY_MAP` only holds Sydney, Brisbane, Melbourne and Newcastle. Anything else falls through to `city`, which is already what `cityStateLine` prints on the line below.

Dropping the fallback lets the venue line disappear when there is nothing real to show. The wrapper on line 264 already handles an empty `venueName`.

```suggestion
  const venueName = event.venue || CITY_MAP[event.city]?.name;
```

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +37,322 @@ type EventsPreviewProps = {
...
+              href={ctaHref}
+              target="_blank"
+              rel="noopener noreferrer"
+              className="block"
+            >
+              <RippleButton className="w-full text-base" variant="primary">
+                {ctaLabel}
+              </RippleButton>
+            </a>
+          )}
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **A button inside a link.** Invalid HTML, and keyboard and screen reader users get odd behaviour out of it.

`RippleButton` renders a real `<button>`, so this comes out as `<a><button></button></a>`. The HTML spec does not allow interactive content inside an `<a>`.

I grepped the repo and this pattern exists nowhere else, so it arrives with this PR.

No clean one-line fix, so it is your call. Either style the anchor itself and drop `RippleButton`, or give `RippleButton` an `asChild` style prop so it renders as the anchor. Same thing happens again in the hero at line 498.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -47,281 +364,266 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
-    if (torsoPresenters.length > 1) return "multi-torso";
-    if (torsoPresenters.length === 1) return "single";
-    if (avatarPresenters.length > 0) return "avatars";
-    return "none";
-  })();
 
-  const showHeaderMedia = resolvedLayout !== "none";
+  // `relativeDate` is "" during SSR then filled on the client, so `isPast`
+  // stays false until hydration and the status line renders consistently.
+  const isPast = relativeDate.endsWith("ago");
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **A finished event shows "Register now" first, then flips to "Watch the recording".**

Open a past event that has a recording. For a moment the button says "Register now" and points at the closed registration page, then it swaps. Google and anyone with JS off only ever see the wrong one.

I can see from the comment that you knew about the SSR gap. But it only covers the status line. `ctaLabel` and `ctaHref` on lines 392 and 393 hang off `isPast` too, and those are the ones that actually matter.

Comparing the end date directly is safe here. Both sides are absolute instants, so no timezone drift and no hydration mismatch.

```suggestion
  // Compare the end date directly so this is right during SSR too.
  const isPast = !!event.endDateTime && new Date(event.endDateTime) < new Date();
```

Bonus: right now `isPast` depends on the wording `"...ago"` coming out of `formatRelativeEventDate`. Reword that copy one day and the recording button silently stops appearing, with nothing to catch it. This kills that coupling too.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -47,281 +364,266 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
+                  )}
+                </div>
+
+                {/* Call to action */}
+                {ctaHref && (
+                  <a href={ctaHref} target="_blank" rel="noopener noreferrer">
+                    <RippleButton className="text-base" variant="primary">
+                      {ctaLabel}
+                    </RippleButton>
+                  </a>
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **Same button-inside-a-link problem as the sidebar CTA.** See my comment on line 291.

`<a>` wrapping `RippleButton`, which renders a `<button>`. Whatever approach you pick for the sidebar, apply it here too so the two stay in step.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -330,47 +632,65 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
+                );
+
+                return url ? (
+                  <CustomLink
+                    key={`presenter-${index}-${name}`}
+                    href={url}
+                    className={cardClass}
+                  >
+                    {cardBody}
+                  </CustomLink>
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **Links inside a link, so the speaker card can stop working.**

`cardBody` renders the speaker bio through `TinaMarkdown`. Plenty of SSW bios have links in them, and an `<a>` inside an `<a>` makes the browser close the outer one early. The card then stops being clickable in the places you would expect.

Suggestion: drop the wrapping link and put the anchor on the speaker's name plus the arrow circle instead. That is where people aim anyway, and it keeps the bio's own links working.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed in 154650b7 — comment now says neutral grey. (You'd already committed the same wording in 4d0886e9; kept yours.)

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed — added `.replace(/\r\n?/g, "\n")` before the newline escape, so a bare CR can't reach the value. Also added RFC 5545 line folding (your other note) — verified the folded output round-trips exactly and no line exceeds 75 chars.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Good catch — took the suggestion. `venueUrl` is now `event.venue ? null : CITY_MAP[event.city]?.url`, so a custom venue renders as plain text rather than linking to the wrong building.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed — dropped the `|| city` fallback as suggested. For a city with no chapel and no venue set, the venue line now disappears instead of printing the city twice.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed — `RippleButton` already had an `href` prop that makes it render a real `<a>` (added for exactly this), so both CTAs now use it instead of being wrapped in an anchor. No more `<a><button>`. I also added `target` to its prop types, which was missing.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Took the suggestion — `isPast` now compares `endDateTime` directly, so SSR and no-JS get the right CTA, and it no longer depends on the wording of `formatRelativeEventDate`. This also resolves your 'finished this morning still reads as Today' note: such an event now reads as past and shows the recording.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed the same way as the sidebar — uses `RippleButton href`, so both stay in step.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed as suggested — the card is a plain `<div>` now, with the link on the speaker's name and the arrow circle only. Bio links inside `TinaMarkdown` no longer nest inside an outer anchor. Gave the arrow an `aria-label` since it's icon-only.

tiagov8:
Thanks @0xharkirat — great review. All 8 inline comments are actioned (replied to each individually), plus the notes:

| Note | Status |
|---|---|
| 13 content files still have `headerLayout` | ✅ Swept all 13 — key is gone from `content/events-calendar/` |
| `dateChip` has the same SSR problem | ✅ Chip renders only once populated, so no empty red square flashes |
| `.ics` has no line folding | ✅ Added RFC 5545 folding (verified: round-trips exactly, no line >75 chars) |
| Event finished this morning reads as "Today" | ✅ Resolved by the `isPast` fix — it now reads as past and shows the recording |
| **Infinite render loop in `useFormatDates`** | ✅ Fixed — the effect now depends on primitive timestamps instead of the `Date` objects the caller rebuilds each render |

The render loop was the big one — thanks for tracing it. It's a pre-existing bug that affects every caller of the hook, not just this page, so the fix is in `hooks/useFormatDates.ts`.

Verified locally: `pnpm lint --max-warnings 0` clean and no new `tsc` errors.

tiagov8:
Two things worth knowing for the re-review:

1. **`hooks/useFormatDates.ts` is shared.** The render-loop fix also affects `upcomingEvents.tsx` and `filter/events.tsx`. It fixes the same pre-existing bug for them, but it's the one change in this PR that reaches beyond the event page.

   While verifying it I found a related latent bug: `startDateTime`/`endDateTime` are optional on `EventTrimmed`, and with missing dates the old code rendered **"today"** (`formatRelativeEventDate` has no guard, so `dayjs(undefined)` resolves to now). The new early return yields an empty string instead. Both other callers just render the string, so this is a strict improvement — flagging it since it's a behaviour change outside this page.

2. **I resolved the review threads as the author.** Each has a reply explaining the fix. If you'd rather verify and close them yourself, reopen any of them — happy to leave that to reviewers next time.

hveraus:
My feedback: 

1. the breadcrumb is not a universal one
2. Move the breadcrumb above the hero
3. the Register now button is not consistent with the others
5. A few elements are off on this 2 buttons; the calendar icon is relatively small, and it is not nice to break the text into 2 lines. Maybe we can ditch the text, the icon does the job. 68d6689f-ad98-4030-867e-c524da9ec0d2
6. the heading for each section and text has different letter spacing, reuse the ones from consulting page 
671c3504-a0ec-4d36-8aa9-54a9f6427a88

tiagov8:

tiagov8:
Thanks @hveraus — all 5 actioned in eb2cd88b.

| # | Feedback | What changed |
|---|---|---|
| 1 | Breadcrumb is not a universal one | Now uses the shared `Breadcrumbs` component (`app/components/breadcrumb`) — the same one `events.tsx` and the consulting pages use. Removed the hand-rolled markup, so it also picks up the mobile back-arrow behaviour for free. |
| 2 | Move it above the hero | Moved — it now sits above the hero card. |
| 3 | Register button inconsistent | It was already `RippleButton`, but I'd styled it `text-base` while the shared `ButtonRow` uses `text-base font-semibold`. Both CTAs now match the standard. |
| 4 | Two buttons are off — small icon, text wrapping | Dropped the labels as you suggested and bumped the icons 16px → 22px, so both fit on one line. Kept `title`/`aria-label` so they're still discoverable for screen readers and on hover. |
| 5 | Inconsistent letter spacing on section headings | Reused the consulting/v3 label style — `font-mono text-sm uppercase tracking-wider text-sswRed` — and applied it to **every** uppercase label on the page (section headings, hero status row, sidebar labels, speaker roles), so they're all one system now. They were previously `font-semibold` + `tracking-widest`, which is what made the spacing look off.

**One thing to decide (#1 side effect):** the shared component builds crumbs from the URL, so on `/events/2026/angularhackdaymelb26` it renders **Home / Events / 2026 / Angular Hack Day - Melbourne**. That `2026` links to `/events/2026`, which isn't a real page. Options:

- leave it (matches how the shared component behaves everywhere), or
- extend `Breadcrumbs` to allow non-linking segments, or
- hide the year crumb on event pages.

Happy to do any of these — just didn't want to fork the shared component's behaviour without a nod. Preview of the current state: https://claude.ai/code/artifact/454bac0d-7284-4961-904c-151d3cba52ba

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -47,337 +394,339 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
-  const city = event.cityOther || event.city;
+  const recordingUrl = event.youTubeId
+    ? `https://www.youtube.com/watch?v=${event.youTubeId}`
+    : event.trailerUrl;
+  const showRecordingCta = isPast && !!recordingUrl;
+
+  // Always state the cost on the button — the price when set, otherwise "Free".
+  const registerLabel = `${event.ctaLabel || "Register now"} · ${
+    event.price || "Free"
+  }`;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **A paid event with no price set advertises itself as free.**

The Tina field description tells the author the opposite:

> Optional price shown on the button, e.g. '$99 AUD'. **Leave blank to hide.**

So someone writes up a $99 workshop, leaves `price` empty because the CMS said blank hides it, and the button ships reading `Register now · Free`. The code comment on line 420 and the schema help text now say opposite things, and the CMS one is the one an editor actually reads.

Pick whichever behaviour you want, but the two need to agree. If blank should hide:

```suggestion
  // Show the cost on the button only when one is set - a blank price hides it,
  // which is what the Tina field description promises the author.
  const registerLabel = event.price
    ? `${event.ctaLabel || "Register now"} · ${event.price}`
    : event.ctaLabel || "Register now";
```

If you would rather keep "Free", then the schema description needs to change instead, and free events should probably set it explicitly.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +33,356 @@ type EventsPreviewProps = {
...
+  ctaHref,
+  schedule,
+}: EventSidebarProps) {
+  const [copied, setCopied] = useState(false);
+
+  const city = event.cityOther || event.city;
+  const state = CITY_MAP[event.city]?.state;
+  // Only link to the SSW chapel page when we're actually using that chapel as
+  // the venue; a custom venue name links nowhere rather than the wrong place.
+  const venueUrl = event.venue ? null : CITY_MAP[event.city]?.url;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **This one is my fault. My suggestion last round was wrong and it broke the normal case.**

I said `venueUrl = event.venue ? null : CITY_MAP[...]`, and you took it as written. It does fix the bug I was aiming at, which was a custom venue linking to the wrong building. But I missed that the schema documents this field with the chapel as its example:

> The specific venue or room, shown before the city in the header, e.g. **'SSW Chapel'**. Leave blank to show the city only.

And the sample event uses exactly that. `CITY_MAP.Melbourne.name` is "SSW Chapel Melbourne". So for the documented, recommended usage the venue is now plain text with no link to sswchapel.com.au. I traded a rare wrong link for a missing link on the common case, which is worse.

What it should have been - link when the venue is the chapel, drop the link when it is somewhere else:

```suggestion
  const chapel = CITY_MAP[event.city];
  const venueUrl =
    !event.venue ||
    chapel?.name?.toLowerCase().startsWith(event.venue.trim().toLowerCase())
      ? chapel?.url
      : null;
```

"SSW Chapel" against "SSW Chapel Melbourne" matches, so the link stays. "Melbourne Convention Centre" does not, so it goes. Worth a quick check on the live page after, since this is the flagship demo event.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -47,337 +394,339 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
+    ? `https://www.youtube.com/watch?v=${event.youTubeId}`
+    : event.trailerUrl;
+  const showRecordingCta = isPast && !!recordingUrl;
+
+  // Always state the cost on the button — the price when set, otherwise "Free".
+  const registerLabel = `${event.ctaLabel || "Register now"} · ${
+    event.price || "Free"
+  }`;
+  const ctaLabel = showRecordingCta ? "Watch the recording" : registerLabel;
+  const ctaHref = showRecordingCta ? recordingUrl : event.url;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **A finished event with no recording still says "Register now" and links to the closed registration page.**

This is the other half of the `isPast` problem I raised last round, and my fix only covered one side of it. `showRecordingCta` needs both `isPast` and a recording. When an event is over but nobody uploaded the video, it falls through to the register branch:

```ts
const ctaLabel = showRecordingCta ? "Watch the recording" : registerLabel;
const ctaHref  = showRecordingCta ? recordingUrl : event.url;
```

So a talk that finished last month shows `Register now · Free` pointing at a dead Eventbrite listing. That is the same misleading state I was trying to remove, reached by a different path.

There are three states here, not two, so the ternaries need a third branch: upcoming, past with a recording, and past without one. For the last, either hide the CTA or swap it for something honest like a link to the events index.

app/components/breadcrumb.tsx
```diff
@@ -42,16 +51,23 @@ const defaultReplacements = [
 
 export const Breadcrumbs: FC<BreadcrumbsProps> = ({
   additionalReplacements = [],
+  excludeSegments = NO_SEGMENTS,
```
0xharkirat, line 56: **Half of this memo fix landed.**

`NO_SEGMENTS` on this line is exactly right, and it is the same class of bug as the render loop, so nice catch. But `additionalReplacements = []` on the line directly above still allocates a fresh array on every render, and both sit in the same `useMemo` dependency list on line 149.

One changing dependency is enough. So the memo still recomputes on every render and the `NO_SEGMENTS` care buys nothing today.

Same treatment fixes it:

```suggestion
  excludeSegments = NO_SEGMENTS,
```

with a matching `const NO_REPLACEMENTS: { from: string; to: string }[] = []` next to `NO_SEGMENTS`, and `additionalReplacements = NO_REPLACEMENTS` above.

Small, but this component renders on 25 pages, so it is worth having the two behave the same.

helpers/dates.ts
```diff
@@ -25,6 +25,66 @@ export const formatEventDate = (start: Date, end: Date) => {
...
+  }
+
+  // Same month: name it once at the end ("Mon 20 - Tue 21 Jul").
+  const dateShort = s.isSame(e, "month")
+    ? `${s.format("ddd D")} - ${e.format("ddd D MMM")}`
+    : `${s.format("ddd D MMM")} - ${e.format("ddd D MMM")}`;
+
+  // "daily" only when the hours actually repeat each day. An event that simply
+  // runs past midnight ends earlier in the day than it starts, and isn't daily.
+  const repeatsDaily = e.format("HH:mm") > s.format("HH:mm");
```
0xharkirat, line 89: **The "daily" label promises something the data cannot back up.**

The type doc on line 36 says `" daily"` appears "when the same hours repeat each day". The check here only compares the start of the first day against the end of the last one, which is the only thing stored.

Take a three-day conference: 9-5, then 10-4, then 9-6. All the data holds is start 9:00 and end 18:00. This returns `9:00 AM - 6:00 PM daily`, which is wrong for two of the three days.

The inference is reasonable and I do not think there is better data to work from. It is the wording that overreaches. Either soften the copy, or soften the doc comment so it describes what the code actually checks rather than what we wish it knew.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -31,11 +33,356 @@ type EventsPreviewProps = {
...
+    .replace(/\\/g, "\\\\")
+    .replace(/;/g, "\\;")
+    .replace(/,/g, "\\,")
+    .replace(/\r\n?/g, "\n")
+    .replace(/\n/g, "\\n");
+}
+
+// Fold content lines longer than 75 octets, per RFC 5545.
+function foldIcsLine(line: string) {
+  if (line.length <= 75) return line;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **The comment says octets, the code counts characters.**

RFC 5545 measures the 75-character limit in octets. `line.length` and `slice()` work in UTF-16 code units, so the two only agree while everything is ASCII.

A presenter named "Tiago Araújo" in the DESCRIPTION already pushes a line past 75 octets before it hits 75 characters. Worse, `slice()` can cut an emoji in half and emit a lone surrogate, which some calendar clients will reject outright.

It needs to measure the thing the comment claims it measures, or the comment needs to stop claiming octets.

hooks/useFormatDates.ts
```diff
@@ -4,37 +4,51 @@ import {
...
+  // effect (and re-render) forever.
+  const startTime = event.startDateTime?.getTime();
+  const endTime = event.endDateTime?.getTime();
 
   useEffect(() => {
-    if (typeof window !== "undefined") {
-      setRelativeDate(
-        formatRelativeEventDate(event.startDateTime, event.endDateTime)
-      );
+    if (startTime == null || endTime == null) return;
```
0xharkirat: **This guard misses the failure mode the callers can actually produce.**

Two small things on this line.

`== null` catches `undefined`, but `mapEventEdge` in `helpers/getTrimmedEvents.ts` always builds `new Date(edge.node.endDateTime)`. Given a bad or absent value that returns an Invalid Date object, never `undefined`. So `getTime()` gives `NaN`, `NaN == null` is false, the guard waves it through, and the page formats "Invalid Date" instead of returning early. `Number.isFinite` is the check that matches what can really arrive.

Second, the early return leaves whatever was formatted last still sitting in state. No path hits it today, since `endDateTime` is `required: true` and every content file has one. Worth clearing anyway if you touch the line.

app/(events)/events/[...filename]/events-preview.tsx
```diff
@@ -47,337 +394,339 @@ export default function EventsPreview({ tinaProps }: EventsPreviewProps) {
...
-    if (avatarPresenters.length > 0) return "avatars";
-    return "none";
-  })();
 
-  const showHeaderMedia = resolvedLayout !== "none";
+  // Compare the end date directly so the CTA/status are correct during SSR too
+  // (both sides are absolute instants — no timezone drift, no relative-copy coupling).
+  const isPast =
+    !!event.endDateTime && new Date(event.endDateTime) < new Date();
+  const statusLabel = isPast ? "Past event" : relativeDate;
```
Comment (0xharkirat) [inline app/(events)/events/[...filename]/events-preview.tsx]: **The comment claims more than the code delivers.**

"the CTA/status are correct during SSR too" is true for the CTA, and true for the status on past events. It is not true for upcoming ones: line 413 falls back to `relativeDate`, which is `""` until the effect runs on the client.

So an upcoming event still renders no status pill on the server and pops one in after hydration. That is fine behaviour and I am not asking you to change it. The comment just needs to stop covering the case it does not cover.

helpers/dates.ts
```diff
@@ -25,6 +25,66 @@ export const formatEventDate = (start: Date, end: Date) => {
...
+
+  return {
+    isMultiDay: true,
+    chips: [chip(s), chip(e)],
+    dateShort,
+    dateLong: `${dateShort} ${e.format("YYYY")}`,
```
0xharkirat: **A multi-day event that crosses New Year loses its start year.**

`dateShort` here has no year on it, and only the end year gets appended. So 31 Dec 2026 to 1 Jan 2027 reads:

```
Wed 31 Dec - Fri 1 Jan 2027
```

which puts the reader on 31 Dec 2027. Rare enough that I would not hold the PR for it, but it is a one-line thing while you are in here: when the two years differ, put the start year on the first half too.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Taken as written — thanks for catching your own one. Now links when the venue *is* the chapel (`"SSW Chapel"` prefix-matches `"SSW Chapel Melbourne"`) and drops the link otherwise, so the documented common case keeps its link and "Melbourne Convention Centre" doesn't get a wrong one.

app/components/breadcrumb.tsx
tiagov8: Right — one moving dependency invalidates the whole memo, so `NO_SEGMENTS` bought nothing on its own. Added `NO_REPLACEMENTS` alongside it and defaulted `additionalReplacements` to it.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed properly rather than softening the comment. It now measures UTF-8 octets via `TextEncoder` and iterates by code point (`for…of`), so nothing splits mid-character. Verified: ASCII, "Tiago Araújo" repeated, and a 60-emoji line all fold to segments of ≤75 octets, round-trip exactly, and contain no lone surrogates.

hooks/useFormatDates.ts
tiagov8: Good catch on the real failure mode — swapped to `Number.isFinite`, since `new Date(bad)` yields NaN rather than undefined. Also clears the four state values on that path so nothing stale is left rendered.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fair — the comment now says the SSR guarantee covers the CTA, and notes upcoming events still wait for hydration before showing a status.

helpers/dates.ts
tiagov8: Fixed. Both halves carry their year when they differ (`Thu 31 Dec 2026 - Fri 1 Jan 2027`), and `dateLong` no longer appends the end year a second time in that case.

helpers/dates.ts
tiagov8: Softened the doc rather than the behaviour, since you're right that there's no better data. It now says " daily" assumes each day keeps the first day's hours, instead of claiming to know they repeat.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Went with your suggestion — blank now hides the suffix, so the code and the Tina description agree and a paid event can never ship advertising itself as free. Free events state it explicitly via the Entry field, which already reads "Free entry" in the hero.

app/(events)/events/[...filename]/events-preview.tsx
Comment (tiagov8) [reply app/(events)/events/[...filename]/events-preview.tsx]: Fixed — there are three branches now. Past with no recording renders no button at all, rather than one pointing at a closed Eventbrite listing. Verified the four states: upcoming → registration, upcoming+price → "Register now · $99", past+recording → "Watch the recording", past without → no button.

tiagov8:
Thanks @0xharkirat — second round all actioned, replied inline on each.

**Fixed as flagged (7)**
- Venue links to the chapel again when the venue *is* the chapel — your earlier suggestion had broken the documented common case.
- `.ics` folding now measures UTF-8 **octets** and splits on code point boundaries. Verified: ASCII, "Tiago Araújo" repeated, and a 60-emoji line all fold to ≤75 octets, round-trip exactly, no lone surrogates.
- `useFormatDates` guards with `Number.isFinite` (Invalid Date ⇒ NaN, not undefined) and clears state on that path.
- `additionalReplacements` got a module-level default too — one moving dep was invalidating the memo on all 25 pages regardless.
- Multi-day events crossing New Year keep both years, without doubling the end year.
- Two comments corrected to stop overclaiming (SSR covers the CTA only; " daily" infers from day one).

**Your call, now decided (2)**
- **Blank price hides the suffix.** Code and CMS text agree; an unpriced paid event can't advertise as free.
- **Past with no recording shows no button.** Three branches now — verified: upcoming → registration, upcoming+price → "Register now · $99", past+recording → "Watch the recording", past without → no button.

All 11 threads resolved. `pnpm lint --max-warnings 0` clean, no new `tsc` errors, 27 tests pass.

0xharkirat (approved):
LGTM

PR #4980, opened 2026-08-12
content: sharpen Claude training page — shared AI vision, two instructors, fresh curriculum

uly1:
## Why

The [Claude training page](https://www.ssw.com.au/consulting/claude-training) undersold several things that matter to a buyer. Every benefit landed on the individual or the pilot group, and three real differentiators were either missing or buried.

## Changes

All in `content/consultingv2/claude-training.json`.

### 1. A shared way of working, not 100 different ones

The org-level outcome was only implied. Now stated:

- **Hero** — appended a sentence naming the outcome ("one shared way of working with AI across your organisation, not 100 different ones")
- **Barrier 03** — retitled *"Good results are hard to repeat"* → *"Every team invents its own way"*, lifting it from team-level variance to the org accumulating competing approaches
- **New Foundation accordion item** — *"A Shared Way of Working with AI"*: agreeing what "good" looks like, safe-use boundaries, shared language, pilot group carrying the standard back
- **New FAQ** — *"Will this give us a consistent approach across the company?"*

### 2. Two instructors, not one

Previously framed as an **Advanced-only** perk, and the cost FAQ used it to justify the Advanced fee. It actually applies to **both tiers**, so:

- **New At a Glance tile** — *"Two Instructors, Not One"* / one leads the teaching, one mentors the room
- **How It Works 02** — Foundation now names both presenters and the teach + mentor split
- **Cost FAQ** — corrected: two presenters are included in each tier's fee, not just Advanced

### 3. Fresh curriculum + industry tailoring

Recency wasn't mentioned anywhere, and tailoring only ever said "business context" — never *industry*:

- **New At a Glance tile** — *"Built Fresh Each Time"* / curriculum designed close to delivery, because AI moves fast
- **Retitled tile** — vague *"One-Stop-Shop"* → *"Tailored to Your Industry"*
- **How It Works 01** — names industry and use-case tailoring, and explains why the curriculum is deliberately designed late

At a Glance goes from 4 tiles to 6, so the two-column grid still fills evenly. All icons (`BiSolidGroup`, `BiSolidBolt`) verified against the Tina icon picker set.

## ⚠️ One thing to sanity-check

The new *"A Shared Way of Working with AI"* accordion item is a claim about what we deliver, not just a copy tweak — it says Foundation includes a "what does good look like for us" alignment exercise. If presenters don't run that today, either the session picks it up or that item needs softening. Everything else here describes what the program already does.

### DIFF
--- a/content/consultingv2/claude-training.json
+++ b/content/consultingv2/claude-training.json
@@ -95,6 +95,11 @@
             "icon": "BiMoneyWithdraw"
           },
           {
+            "heading": "Two Instructors, Not One",
+            "description": "One leads the teaching, one mentors the room, so nobody gets stuck",
+            "icon": "BiSolidGroup"
+          },
+          {
             "heading": "Hands-On by Design",
             "description": "Real prompts, workflows, skills and live prototypes",
             "icon": "BiSolidBookBookmark"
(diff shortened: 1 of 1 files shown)


miichaelsmedley (approved):
Perfect

PR #4984, opened 2026-08-13
💄 Redesign /products index + third-party brand card tokens

hveraus:
Redesigns the `/products` index, which had drifted well behind the rest of the
site: it reused the generic `PageCard` in a two-column list, so every product —
first-party and third-party alike — got identical chrome, and the page had no
dark mode while `/` and `/consulting` did.

This brings it in line with the `/consulting` index redesigned in #4939: the
same themed shell, the same card language, and a four-column grid that actually
uses the width. Products now carry short capability tags so the grid can be
skimmed without reading every description.

- Affected routes: `/products`

## What changed

**The page** — [products-index.tsx](app/(about)/products/products-index.tsx)
- Wrapped in `HomeThemeShell` and added `/products` to `THEMED_ROUTES`, so the
  page gets the same dark theme (and pre-paint script) as `/` and `/consulting`.
- Replaced the 1/2-column list with a responsive 1 → 2 → 4 column grid.
- The `subTitle` heading is gone — it was generic marketing copy above a grid
  that now explains itself. Removed from the Tina collection and the content
  JSON too, not just hidden.
- The component now takes `data` rather than an untyped `props`, typed as
  `ProductsIndexQueryQuery`. `products-index-preview.tsx` updated to match.
- Fixed the title's visual-editing attribute: `props-tina-field` was never a
  real attribute, so the `h1` never registered as a click target in Tina.

**The cards** — new [components/products/](components/products/)
- `productCard.tsx` — the standard product card.
- `brandProductCard.tsx` — one config-driven component for the two products
  whose media kits need their own surface (TinaCMS, YakShaver). Each brand is a
  `BrandCardConfig`; only artwork and brand classes differ. These render
  double-width from `md` up and are pinned to the front of the grid by a stable
  sort, so CMS reordering doesn't move them.
- `productCardShell.tsx` — `CustomLink` returns a bare fragment when `href` is
  falsy, which dropped the element carrying the card chrome and leaked the
  card's children into the grid as loose items. `url` is optional on
  `productsList`, so this falls back to a plain `div`. Covered by tests.
- `shared.ts` — chrome shared across all three card kinds, plus `visibleTags`
  (trim → drop blanks → dedupe → cap at 2) and `destinationLabel`.

**Content and schema**
- New `tags` field on `productsList` (`list: true` strings) —
  [tina/collections/products.tsx](tina/collections/products.tsx) and
  [index.json](content/products/index/index.json). Every product gets two,
  except SSW Rewards which has three — only the first two render, so it also
  serves as the live case for the cap.
- Removed the `subTitle` field.
- Reordered the list: SSW Dory moves up, SmashingBarrier moves to last.
- Tightened the TinaCMS, SophieBot, SophieHub and SmashingBarrier descriptions.
- `destinationLabel` resolves against a base URL because SSW Rewards' `url` is
  a relative `/products/rewards`.

**Tokens** — [styles.css](styles.css), [tailwind.config.js](tailwind.config.js)
- Two TinaCMS brand colours, namespaced `--brand-tina-*` so they stay visibly
  separate from the SSW palette. Both are darkened from the published brand
  orange to clear 4.5:1 with white text; the measurements are in the comment.

**Tests** — [__tests__/components/products/productCard.tsx](__tests__/components/products/productCard.tsx)
- 8 tests over the card shell (the no-`url` regression above) and tag handling.
- `jest.config.ts` gains an `^@/lib/(.*)$` alias — `tsconfig` maps a catch-all
  `@/*`, and the component tests need the part of it that resolves `cn()` from
  `@/lib/utils`.

**Unrelated local-dev fix** — [next.config.mjs](next.config.mjs)
- Pinned `turbopack.root`. A stray lockfile in a parent directory (e.g.
  `~/package-lock.json`) makes Turbopack infer the workspace root above the
  project, which breaks resolution of root-level files like
  `instrumentation.ts`. Not related to `/products`, but it blocked local dev on
  this branch.

## Assets
- `public/images/YakShaver-Gradient-BG.png` — the YakShaver card background.
- `public/images/company-logos/TinaCms-Logo-Full-Default.svg` — the TinaCMS
  lockup, from their media kit.


hveraus:

components/products/shared.ts
```diff
@@ -0,0 +1,170 @@
+// Shared chrome for the three kinds of card on /products (standard, TinaCMS,
```
isaaclombardssw: **Comment volume.** `components/products/` is 477 of 888 lines comment (54%), and this file is 127 of 170 (74%). The baseline across `components/` is about 4%.

The twMerge traps are worth keeping - I ran twMerge and confirmed `outline-2` really does swallow the bare `outline` and silently kill the focus indicator. That is a genuine trap and the comment earns its place.

What I would cut is the revision narrative ("an earlier revision lifted the card 6px", "two intermediate revisions were still too strong") which git already stores, and the hand-measured contrast ratios, which nothing tests and which quietly become false the moment someone changes `--card`.

components/products/productCard.tsx
```diff
@@ -0,0 +1,192 @@
...
+// their own components instead, because their media kits require their own
+// surfaces.
+export const ProductCard: FC<ProductCardProps> = ({ product, tinaNode }) => {
+  const domain = destinationLabel(product.url);
+  const tags = visibleTags(product.tags);
+
+  return (
+    // ProductCardShell, not a raw <a>: for a product that has a url it wraps
+    // CustomLink, which routes the one on-site product (SSW Rewards, at
+    // https://www.ssw.com.au/products/rewards) through next/link and adds
```
isaaclombardssw: This says `https://www.ssw.com.au/products/rewards`, but the content has no `www` - it is `https://ssw.com.au/products/rewards`. The missing `www` is exactly what causes the redirect hop flagged in `index.json`.

A second one at line 49: the chrome is not identical to `ConsultingCard`, which uses `rounded-xl` where this uses `rounded-card`. Two comments drifted from the code inside one PR is the argument for fewer of them.

content/products/index/index.json
```diff
@@ -10,68 +10,113 @@
...
-    {
-      "name": "SmashingBarrier",
-      "url": "https://smashingbarrier.com/",
-      "description": "SmashingBarrier is an user control that integrates with web applications, allowing users to send feature requests or issue reports directly into related team projects.",
-      "logo": "/images/products/thumb-SmashingBarrier.jpg"
-    },
     {
       "name": "SSW Rewards",
-      "url": "/products/rewards",
+      "url": "https://ssw.com.au/products/rewards",
```
isaaclombardssw: **Please put the relative path back.** This was `/products/rewards` on `main`.

`customLink.tsx` still classifies it as internal (it contains `ssw.com.au` and is not a `/people`, `/rules` or `/ssw` URL), so it does render `next/link` - but `next/link` cannot soft-navigate or prefetch a cross-origin absolute URL, so this is now a full page load. `curl` also confirms that host 301s to `www`, so it is a full load plus a redirect hop.

The bit that worries me more: on a PR preview slot this walks the reviewer off the preview and into production.

tsconfig.json
```diff
@@ -1,31 +1,55 @@
...
+    ],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": false,
     "forceConsistentCasingInFileNames": true,
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
isaaclombardssw: **Please revert this file, or split it out with the reasoning stated.**

`prettier --check tsconfig.json` passes on `origin/main` and fails on this branch. Nothing catches it, because `pnpm lint` is `eslint . --ext .ts,.tsx` so JSON never reaches the formatter.

The `moduleResolution` node-to-bundler and `jsx` preserve-to-react-jsx flips, the array reformat and the `.next/dev/types` include are all Next's own `writeConfigurationDefaults` output (`node_modules/next/dist/lib/typescript/writeConfigurationDefaults.js:151,180`) - a dev-server auto-rewrite that got committed. It may well be harmless, and CI is green, but it changes type resolution repo-wide and it is not mentioned in the description.

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,135 @@
...
-            {props.productsIndex.subTitle}
-          </h2>
-        )}
-        <div className="flex flex-col md:flex-row">
-          <div className="grid w-full grid-cols-1 gap-2 lg:grid-cols-2">
-            {props.productsIndex.productsList?.map((product, index) => (
-              <PageCard page={product} key={index} />
-            ))}
-          </div>
+        {/* The CMS `subTitle` is deliberately not rendered. It still holds
```
isaaclombardssw: **Please remove `subTitle` properly rather than dropping it at the render.**

The reasoning here is that the field is `required: true` so it cannot be emptied - but this PR already edits `tina/collections/products.tsx` to add `tags` and regenerates `tina-lock.json`, so the collection is in play.

As it stands an editor sees a mandatory field, fills it in, and nothing happens on the page, with no signal anywhere in the CMS. Either drop `required`, or add a `description` saying it is not displayed - exactly what the new `tags` field does.

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,135 @@
...
+const isBrandCard = (name?: string) =>
+  BRAND_CARDS.includes((name ?? "").trim().toLowerCase());
+
 export default function ProductsIndexContent({ props }) {
+  const node = props.productsIndex;
+  const products = node?.productsList ?? [];
+
+  // Each brand card occupies two grid cells at the tiers where it spans, so the
+  // trailing gap the panel fills is measured in cells, not products.
+  const cellCount =
```
isaaclombardssw: **Worth making the ordering explicit.** This is exact only because TinaCMS and YakShaver happen to lead the CMS list.

A `col-span-2` card that cannot fit the remaining columns is bumped to the next row and leaves the rest of the current row empty (sparse auto-placement, the default). `cellCount` never sees that hole. I simulated placement across orderings: with the current order the panel span is right at both tiers, but move any standard card in front of a brand card and real cells go 13 to 14, so at 4-up the panel computes span 3 into 2 free slots, cannot fit, and wraps to a row of its own - the ragged trailing row the panel exists to prevent.

Sorting brand cards to the front before the map is one line, makes this count provably exact (2 divides both 2 and 4, so brand-first never leaves a hole), and changes nothing visually today.

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,135 @@
...
+import { TinaProductCard } from "@/components/products/tinaProductCard";
+import { YakShaverProductCard } from "@/components/products/yakShaverProductCard";
 import { Container } from "@/components/util/container";
 import { Breadcrumbs } from "app/components/breadcrumb";
 import { tinaField } from "tinacms/dist/react";
 
+// The two products whose owners' media kits require their own card surface.
+// Matched on name rather than list position so the treatment follows the
+// product if an editor reorders the CMS list.
+const BRAND_CARDS = ["tinacms", "yakshaver"];
```
isaaclombardssw: Minor: this list drives `cellCount`, but rendering branches on hardcoded `name === "tinacms"` / `"yakshaver"` literals at lines 106 and 114. Hardcoding these two is completely fine - it is just stated in two places, so adding a third name here would inflate the panel span against a card that never actually spans.

One `Record<string, FC>` that both the count and the render read from would close it.

content/products/index/index.json
```diff
@@ -10,68 +10,113 @@
...
       "description": "Scan SSW QR codes, earn SSW Points, claim rewards, and win prizes!",
+      "tags": [
+        "Mobile App",
+        "Fun Quizes",
```
isaaclombardssw: Typo: "Quizes" should be "Quizzes". Currently invisible because `MAX_VISIBLE_TAGS` is 2 and this is the third tag, which is presumably why it slipped through.

isaaclombardssw (changes requested):
## 🤖 Agent Review

Requesting changes on four things, all commented inline against the diff.

### 🚨 Must Fix / Address
- 💄 **Comment volume** - 477 of 888 lines in `components/products/` are comment (54%), and two of them are already wrong.
- **SSW Rewards link** - relative to absolute non-www, which costs the soft navigation and adds a 301 hop.
- **tsconfig.json** - no longer Prettier-clean where `main` was, plus an undeclared repo-wide `moduleResolution`/`jsx` flip.
- **subTitle** - still `required: true` and populated, but never rendered.

### 🤏 Small Feedback
- Make the brand-card ordering explicit rather than assumed.
- `BRAND_CARDS` and the render branches state the same list twice.
- "Fun Quizes" typo in the Rewards tags.

### 📞 Callouts
- **The rest is in good shape** - CI passes on this SHA, tests and lint are clean locally, and the `THEMED_ROUTES` opt-in, the `data-tina-field` fix and the falsy-href guard in `ProductCardShell` are all nicely done. This is a request for changes on the list above only.

components/products/shared.ts
hveraus: done with Claude:

Removed revision narrative — sentences describing prior iterations (e.g. "an earlier revision lifted the card 6px," "two intermediate revisions were still too strong," the removed install-command chip / gradient blend-mode history). Git history already covers this.
Removed hardcoded contrast ratios — every manually-measured X.XX:1 value, since nothing tests them and they'd silently go stale if a --card/--brand token changes. Kept the accessibility intent instead (e.g. "clears WCAG 3:1," "would fail AA in dark mode").
Kept as-is: the outline-2/twMerge focus-indicator warning in shared.ts, plus all other why-comments explaining twMerge conflicts, library quirks, and non-obvious trade-offs.

Net diff: -142/+103 lines, comment-only (verified no non-comment lines touched), tests still pass. Also fixed a stale doc-comment in yakShaverProductCard.tsx that claimed the text column was width-capped when the actual code (and the per-element comment) shows it isn't.

components/products/productCard.tsx
hveraus: Addressed the comment and made the requested display change:

Fixed the stale/inaccurate comments they flagged (productCard.tsx):

The doc-comment referenced https://www.ssw.com.au/products/rewards — updated to https://ssw.com.au/products/rewards to match the actual content (no www).
The comment claiming ConsultingCard and this card have "identical card chrome" was wrong — ConsultingCard uses rounded-xl (12px), this uses rounded-card (16px). Reworded to state parity is on colour/motion only, not shape, so the comment can't drift from the code again.
Changed the displayed URL (shared.ts):

destinationLabel used to strip the path and show hostname only. It now keeps the path when the URL points below the domain root, so SSW Rewards' footer shows ssw.com.au/products/rewards instead of the ambiguous ssw.com.au (which read like a link to the homepage).
Side effect, not a regression: SophieBot's URL also has a non-root path (/sophiebot), so its footer label picked up the same fix and now shows sswsophie.com/sophiebot instead of just sswsophie.com. Every other product's URL is root-path, so their labels are unchanged.

content/products/index/index.json
hveraus: done

tsconfig.json
hveraus: done

app/(about)/products/products-index.tsx
hveraus: Don't understand, but this is what Claude fixed:

Removed subTitle from the schema properly rather than leaving it dropped at render:

tina/collections/products.tsx — deleted the field definition entirely.
content/products/index/index.json — removed the orphaned subTitle key and its content.
app/(about)/products/products-index.tsx — removed the now-obsolete comment explaining the render-time drop.
tina/tina-lock.json — regenerated via tinacms build --content=local. Verified productsIndex now has ['seo', 'title', 'productsList'] (matches main minus subTitle), productsList still carries tags, and the total collection count is unchanged.

app/(about)/products/products-index.tsx
hveraus: Fixed the correctness bug the reviewer flagged in app/(about)/products/products-index.tsx: cellCount was only accurate because TinaCMS and YakShaver happen to lead the CMS list today. Under CSS Grid's default sparse auto-placement, a col-span-2 brand card that doesn't fit the remaining columns of its row gets bumped to a fresh row, leaving the skipped cell empty — cellCount's arithmetic can't see that gap, so a reorder in the CMS would make the trailing MoreProductsPanel compute the wrong span and wrap into its own ragged row.

Fix: sort brand cards to the front of the render order before computing cellCount and mapping the grid, using a stable sort so CMS order is otherwise preserved on both sides of the split. Verified this makes the count provably correct (2 divides both the 2-col and 4-col breakpoints, so brand-first can never leave a gap) and changes nothing visually today — confirmed the live-rendered grid order is exactly TinaCMS, YakShaver, SugarLearning, EagleEye, TimePro, SophieBot, SophieHub, SSW Dory, CodeAuditor, SSW Rewards, SmashingBarrier, unchanged from the CMS order.

Bonus catch along the way: while verifying this, I hit a page-breaking error (Cannot read properties of null (reading 'data')) that turned out to be unrelated to this edit — the manual tinacms build I ran earlier for the subTitle fix had left the running dev server's Turbopack/Tina cache stale. Traced it by querying the local GraphQL server directly and calling the generated query function standalone (both succeeded, proving the schema/content were fine), then restarted the dev server to clear the stale cache. It's now healthy and serving the page correctly.

app/(about)/products/products-index.tsx
hveraus: Consolidated the brand-card lookup into a single source of truth, per the reviewer's suggestion, since the previous BRAND_CARDS array plus the two hardcoded name === "tinacms" / name === "yakshaver" render branches were three separately-maintained lists that cellCount and the JSX could silently drift apart from.

Change in app/(about)/products/products-index.tsx:

Replaced BRAND_CARDS/isBrandCard with BRAND_CARD_COMPONENTS: Record<string, FC<BrandCardProps>> mapping "tinacms" → TinaProductCard and "yakshaver" → YakShaverProductCard, plus a brandCardFor(name) lookup.
The sort (brand-first ordering), cellCount, and the render's brand-card branch all now read from this one map — the two if (name === ...) branches collapsed into a single if (BrandCard) return <BrandCard .../>.
Kept it as a plain object rather than reaching for anything fancier, since it's still just two entries.

content/products/index/index.json
hveraus: done

hveraus:

isaaclombardssw (changes requested):
**Please cut the comments right down - this is the one item from last round that measurably didn't move.** `components/products/` went from 40.1% comment lines to 39.6%, and `shared.ts` is still 119 comment lines to 34 of code. The contrast ratios moved to `styles.css` rather than going away. `components/consulting/` next door is 7 comment lines across 117 lines of code.

About 10 lines across the whole folder is what I'd expect. Worth keeping: the `border-0.75` note (`shared.ts:9`), the twMerge `outline-2` trap (`shared.ts:26`), why the spans are looked up in a map instead of interpolated, and the contrast values in `styles.css`. The rest is restating the code, aesthetic rationale that belongs in the PR description, or revision narrative git already stores.

The risk isn't just noise. Nothing compiles or tests a comment, so volume is exposure - and two are wrong right now. `shared.ts:144` says every product is authored with an absolute URL, which your own Rewards fix falsified this round, and it sits directly above the `new URL(url, base)` fallback telling the next person the base isn't needed. `shared.ts:136` lists duplicates as a case `visibleTags` handles, but it doesn't, and `key={tag}` throws on a repeated tag.

hveraus:
> **Please cut the comments right down - this is the one item from last round that measurably didn't move.** `components/products/` went from 40.1% comment lines to 39.6%, and `shared.ts` is still 119 comment lines to 34 of code. The contrast ratios moved to `styles.css` rather than going away. `components/consulting/` next door is 7 comment lines across 117 lines of code.
> 
> About 10 lines across the whole folder is what I'd expect. Worth keeping: the `border-0.75` note (`shared.ts:9`), the twMerge `outline-2` trap (`shared.ts:26`), why the spans are looked up in a map instead of interpolated, and the contrast values in `styles.css`. The rest is restating the code, aesthetic rationale that belongs in the PR description, or revision narrative git already stores.
> 
> The risk isn't just noise. Nothing compiles or tests a comment, so volume is exposure - and two are wrong right now. `shared.ts:144` says every product is authored with an absolute URL, which your own Rewards fix falsified this round, and it sits directly above the `new URL(url, base)` fallback telling the next person the base isn't needed. `shared.ts:136` lists duplicates as a case `visibleTags` handles, but it doesn't, and `key={tag}` throws on a repeated tag.

Comment volume: cut components/products/ from 282 comment lines (35.6%) down to 27 (6.8%) — 433 lines removed net. Kept only the load-bearing notes: the border-0.75 rationale and the outline/twMerge outline-2 trap in shared.ts, plus a few genuine invisible-breakage guards elsewhere (the CustomLink fragment-collapse reason in productCardShell.tsx, the dark-scope corollaries and pr-10-must-follow-p-6 twMerge trap on the two brand cards, and why the brand cards' titles are sr-only). Everything aesthetic/rationale/narrative was deleted. Left styles.css's contrast-ratio comments untouched, as the reviewer said to keep those.

The two wrong comments, both fixed at the code level, not just reworded:

shared.ts no longer claims every product has an absolute URL — SSW Rewards' /products/rewards is relative, which is exactly why destinationLabel takes a base URL; the comment now says that plainly.
visibleTags now actually dedupes tags (Array.from(new Set(...))) instead of just claiming to, which also fixes the key={tag} collision risk in the three card components since tags rendered are now guaranteed unique.
Added a regression test (drops duplicate tags an editor left behind) covering the dedupe fix. Lint and the full productCard test suite pass (8/8).

__tests__/components/products/productCard.tsx
```diff
@@ -0,0 +1,146 @@
+import { ProductCard } from "@/components/products/productCard";
+import { render, screen } from "@testing-library/react";
+// This repo leaves setupFilesAfterEnv commented out in jest.config.ts, so the
+// jest-dom matchers are not registered globally. Import them here.
+import "@testing-library/jest-dom";
+
+// tinacms/dist/react is published as ESM and Jest does not transform
```
joshbermanssw, line 7: Dont think we need this file? __tests__/* anything? 

Should it be .gitignored

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,131 @@
...
+  const products = [...(node?.productsList ?? [])].sort(
+    (a, b) => Number(!!brandCardFor(b?.name)) - Number(!!brandCardFor(a?.name))
+  );
+
   return (
-    <>
-      <Container className="mb-10 flex-1 pt-2">
-        <Breadcrumbs path={"/products"} title={"Products"} />
-        {props.productsIndex.title && (
+    // min-h-screen, not min-h-full: PageLayout's <main> carries an
```
joshbermanssw: Reduce all comments on this file (if possible remove)

public/images/company-logos/TinaCms-Logo-Full-Default.svg
```diff
@@ -0,0 +1,11 @@
+<svg width="1020" height="254" viewBox="0 0 1020 254" fill="none" xmlns="http://www.w3.org/2000/svg">
```
joshbermanssw, line 1: pls dont write out the svg code

styles.css
```diff
@@ -125,6 +125,35 @@
...
+       THIRD-PARTY BRAND COLOURS — not part of the SSW palette.
+       These are TinaCMS's and YakShaver's own brand values, taken from their
+       official media kits (tina.io/media, yakshaver.ai/media). They are kept
+       in this separate --brand-<company>-* namespace so they are never
+       mistaken for SSW tokens and never reused as generic UI colours.
+       Declared once, in :root only — deliberately NOT overridden in .dark,
+       because another company's brand colour does not follow our theme.
+       --------------------------------------------------------------------- */
+
+    /* Official Tina orange. Same value as public/images/megamenu-icons/
```
joshbermanssw: remove comments

public/images/YakShaver-Gradient-BG.png
joshbermanssw, line 1: pretty sure u can get the gradient in code from SSW.Products

joshbermanssw (commented):
couple things

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,131 @@
...
+            if (BrandCard) {
+              return (
+                <div key={key} className="md:col-span-2">
+                  <BrandCard product={product} tinaNode={product} />
+                </div>
+              );
+            }
+
+            return (
+              <ProductCard key={key} product={product} tinaNode={product} />
```
joshbermanssw: **A UI component should never take a `tinaNode`.**

Both call sites — this line and :118 — pass `product={product} tinaNode={product}`. The same object, twice, under two names, the second typed `Record<string, unknown>`.

That prop exists only because the hand-written `product` type discards Tina's generated type *and* the `_content_source` metadata `tinaField` reads off the object. So a second escape-hatch prop was needed to hand that metadata back. It's `any` with extra steps.

`product` **is** the Tina node — that's precisely why both props receive the same value. Type `product` from the generated type and call `tinaField(product, "name")` directly, the way `app/articles/index.tsx:56` already does.

What goes away: the prop, its type, 18 references, the 9 copies of `tinaNode ? tinaField(tinaNode, "x") : undefined` guarding a prop no call site omits, and three of the four duplicate shape types.

```tsx
<ProductCard key={key} product={product} />
```

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,131 @@
...
+const BRAND_CARD_COMPONENTS: Record<string, FC<BrandCardProps>> = {
+  tinacms: TinaProductCard,
+  yakshaver: YakShaverProductCard,
+};
+
+const brandCardFor = (name?: string) =>
+  BRAND_CARD_COMPONENTS[(name ?? "").trim().toLowerCase()];
+
 export default function ProductsIndexContent({ props }) {
+  const node = props.productsIndex;
```
joshbermanssw: Same point at the entry: this page renders products, not nodes. `node` is CMS vocabulary and it's the first thing the file says.

Related, on the line above: `export default function ProductsIndexContent({ props })` has an untyped parameter, so it's implicitly `any` under `strict: false` — and the prop is literally named `props`, so this reads `props.props`. Every access below it (`props.productsIndex`, `node?.title`, `node?.productsList`) is unchecked, and a field rename in the collection fails silently at runtime instead of at build.

app/(about)/products/products-index.tsx
```diff
@@ -1,37 +1,131 @@
...
+    description?: string;
+    tags?: string[];
+  };
+  tinaNode?: Record<string, unknown>;
+};
+
+// The two products whose owners' media kits require their own card surface,
+// keyed by lowercase name so the treatment follows the product if an editor
+// reorders the CMS list.
+const BRAND_CARD_COMPONENTS: Record<string, FC<BrandCardProps>> = {
```
joshbermanssw: `Record<string, FC<BrandCardProps>>` throws away the keys — `brandCardFor` returns `FC<BrandCardProps> | undefined` and nothing knows the map only holds `"tinacms"` and `"yakshaver"`.

`satisfies` keeps both the check and the inference:

```ts
const BRAND_CARD_COMPONENTS = {
  tinacms: TinaProductCard,
  yakshaver: YakShaverProductCard,
} satisfies Record<string, FC<BrandCardProps>>;
```

components/products/yakShaverProductCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+    url?: string;
+    description?: string;
+    tags?: string[];
+  };
+  tinaNode?: Record<string, unknown>;
+};
+
+// YakShaver's card: their dark-mode lockup and brand gradient artwork on the
+// same surface the other nine products use in dark mode.
+export const YakShaverProductCard: FC<YakShaverProductCardProps> = ({
```
joshbermanssw: This and `tinaProductCard.tsx` are the same component.

Normalised for names, the two files differ by roughly 15 lines across 112 and 103. Identical shell, identical `sr-only` h3 carrying the name, identical tags `ul`, identical "Learn More" span, identical `ProductCardShell` wrapper.

The actual difference is two image paths, one class string and one tag-chip token — that's a brand config object, not a second file. As it stands any fix to the shared structure has to be made twice and nothing catches it when it isn't.

components/products/tinaProductCard.tsx
```diff
@@ -0,0 +1,103 @@
...
+      className={cn(
+        cardShell,
+        "justify-between gap-4 bg-brand-tina-field p-6",
+        // `dark` scopes every token in this card to its dark value, since the
+        // orange field doesn't follow the page theme.
+        "dark",
+        "border-hairline hover:border-brand",
+        "hover:bg-brand-tina-field-hover",
+        "focus-visible:outline-white",
+        // Must stay after p-6: twMerge lets a later p-* override an earlier
```
joshbermanssw: This comment is copy-pasted verbatim to `yakShaverProductCard.tsx:49`.

It documents a real ordering constraint — `pr-10` must follow `p-6` or twMerge drops it — but nothing enforces it. Reorder the classes in either file and the padding silently disappears, in both places, with the comment still sitting there saying it won't.

A constraint that matters enough to document twice is worth encoding once: compose the padding into a single token in `shared.ts`, or pin it with a test. The duplicated paragraph is the weakest of the three options.

components/products/productCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { ArrowCircle } from "../blocks/v3/shared/arrowCircle";
+import {
+  cardShell,
+  destinationLabel,
+  productTagChip,
+  visibleTags,
+} from "./shared";
+import { ProductCardShell } from "./productCardShell";
+
+export type ProductCardProps = {
```
joshbermanssw, line 16: Fourth hand-written copy of the same shape — alongside `BrandCardProps` (`products-index.tsx:10`), `TinaProductCardProps` (`tinaProductCard.tsx:11`) and `YakShaverProductCardProps` (`yakShaverProductCard.tsx:19`).

All four restate a shape Tina already generates from the collection. Add a field to `productsList` and you edit four types by hand, and nothing fails if you miss one. Derive them from the generated type instead.

joshbermanssw (dismissed):
## Code quality

The visuals are fine. The problem is that this page is written in the CMS's vocabulary instead of the domain's, and it's hard to read as a result.

`/products` is a grid of products. The only noun it should need is `product`. Right now `products-index.tsx` says "node" twice before it says anything about a product, and every card takes a `tinaNode` prop typed `Record<string, unknown>`. That's CMS transport plumbing leaking into the view layer, and it's the root of most of what follows — four hand-written copies of one shape, 9 defensive ternaries, and an untyped entry point. Fix the types at the boundary and the rest collapses on its own.

Separately: **the PR description no longer matches the branch.** It still describes `MoreProductsPanel` and sizing it "from the occupied cell count rather than hardcoded" — both were deleted in b725e3b7 and 3e8135c4 and neither exists in the code. The template placeholders (`- Affected routes:`, `Fixed #{{ ISSUE NUMBER }}`) are also still unfilled.

__tests__/components/products/productCard.tsx
hveraus: It's intentional, not stray — `__tests__` is already an established, tracked pattern in this repo (e.g. `__tests__/components/v2ComponentWrapper.test.tsx`, `__tests__/helpers/consultingSections.ts` from earlier PRs), and `.gitignore` only excludes `/test-results/` (Jest's output), not `__tests__` (the source).

CI actually runs it: `pr-tests.yml` runs `pnpm test-and-coverage` (Jest) on every PR into `main` and posts a coverage report, so these need to be committed to be picked up.

This particular file is a regression guard, not incidental scaffolding — it pins the fix for a real bug where `CustomLink` drops the card's chrome classes when a Tina-authored product has no `url`, so the children get emitted as loose siblings in the grid instead of inside one card. Without the test, that could silently come back on a future refactor.

app/(about)/products/products-index.tsx
hveraus: Trimmed in 2b7bc51df — cut the long design-rationale blocks down to short one-liners, kept only the ones guarding a real gotcha (dark-mode white band, the 18px h1 margin bug, horizontal scroll from col-span-2 at the base grid tier).

public/images/company-logos/TinaCms-Logo-Full-Default.svg
hveraus: This is the real TinaCMS asset, not hand-typed/generated code — the icon paths match TinaCMS's own npm package (`@tinacms/app/favicon.svg`) scaled up, and the 7 wordmark paths after it decode letter-by-letter into "t-i-n-a-c-m-s". SVG is just XML text, so any addition to this file shows as a markup diff in review regardless of source — there's no way to add an SVG without it appearing as "written out" code.

styles.css
hveraus: done.

public/images/YakShaver-Gradient-BG.png
hveraus: tried, but hard to get the size and angle I want

app/(about)/products/products-index.tsx
hveraus: Done — product is now typed ProductsIndexProductsList (from @/tina/types) everywhere, and tinaNode is gone entirely: the prop, its type, all 18 references, and the 9 guarded tinaNode ? tinaField(tinaNode, "x") : undefined ternaries. tinaField(product, "...") is called directly now, same as app/articles/index.tsx:56.

Changed: products-index.tsx, productCard.tsx, tinaProductCard.tsx, yakShaverProductCard.tsx (types + call sites), plus the two <BrandCard>/<ProductCard> call sites that were passing the same object twice.

One behavioural note: tinaField returns "" (not undefined) when _content_source metadata is absent, so data-tina-field now renders as data-tina-field="" in that case instead of being omitted — consistent with how articles/index.tsx already behaves, not a regression.

app/(about)/products/products-index.tsx
hveraus: Fixed — ProductsIndexContent now takes { data }: { data: ProductsIndexQueryQuery } (the actual generated type from client.queries.productsIndexQuery, imported from @/tina/types), and node is renamed to productsIndex throughout. Updated the one call site (products-index-preview.tsx) to pass data={data} instead of props={data}.

Field accesses (productsIndex.title, productsIndex.productsList) are now type-checked against the real Tina schema, so a collection field rename would fail at build time, not silently at runtime.

Flagging one thing out of scope for this comment: products-index-preview.tsx's own { props } param has the identical untyped-props smell one layer up, and consulting/index.tsx still uses node for the same pattern — happy to do those as a follow-up if you'd like.

app/(about)/products/products-index.tsx
hveraus: Applied exactly as suggested — BRAND_CARD_COMPONENTS now infers its literal key type ("tinacms" | "yakshaver") via satisfies Record<string, FC<BrandCardProps>> instead of widening to Record<string, FC<BrandCardProps>> and throwing the keys away. Reply draft:

Done — switched to satisfies exactly as you suggested, so the object keeps its literal key inference while still being checked against the Record<string, FC<BrandCardProps>> shape.

components/products/yakShaverProductCard.tsx
hveraus: Done — tinaProductCard.tsx and yakShaverProductCard.tsx are gone, replaced by one components/products/brandProductCard.tsx with two config objects (TINA_CARD_CONFIG, YAKSHAVER_CARD_CONFIG). To be sure I didn't introduce any visual drift, I wrote a throwaway test rendering the old vs. new components with identical props and diffed the HTML — it caught a real regression on the first pass (I'd wrapped YakShaver's content in an extra flex div that wasn't there before, changing its gap spacing), which I fixed by making the content wrapper config-driven per brand. Final diff confirmed byte-identical markup for both cards, then I deleted the scratch test.

Merged into a single components/products/brandProductCard.tsx, config-driven (TINA_CARD_CONFIG / YAKSHAVER_CARD_CONFIG). products-index.tsx now renders <BrandProductCard product={product} config={brandConfigFor(product.name)} /> instead of picking between two components.

Verified with a throwaway diff test (old component vs. new + config, same props, HTML compared) before deleting the two files — output is unchanged for both brands, including the no-url/no-description/no-tags edge cases.

components/products/tinaProductCard.tsx
hveraus: Fixed. That duplication is already gone — tinaProductCard.tsx/yakShaverProductCard.tsx were merged into a single brandProductCard.tsx in an earlier round, so this comment now exists exactly once, not twice.

Your deeper point still applied to that one copy though: nothing stopped someone from splitting p-6/pr-10 back into two reorderable strings later. Composed it into a single brandCardPadding token in shared.ts per your suggestion, so the ordering can't be silently broken by a future edit — verified the rendered class list still ends ...p-6 pr-10 (right override intact) with a quick render check.

components/products/productCard.tsx
hveraus: Already resolved by the earlier fixes in this thread — down from four hand-written shapes to two, and both (ProductCardProps, BrandProductCardProps) are { product: ProductsIndexProductsList }, deriving from the generated type rather than restating it. A field added to productsList now flows through both automatically; nothing left to hand-edit or silently miss.

hveraus:

components/products/shared.ts
isaaclombardssw: Confirmed, closing this. `components/products/` is now 33 comment lines across ~350 of code (~9%), down from 40%. `shared.ts` is 13/56 and each one earns its place — the `border-0.75` note, the twMerge `outline-2` trap, the `p-6 pr-10` ordering.

Both of the comments that were *wrong* last round are fixed rather than just deleted: `shared.ts:46` now says the opposite (correctly — `url` can be relative), and `visibleTags` genuinely dedupes now, with a test pinning it.

components/products/productCard.tsx
isaaclombardssw: Confirmed. Both stale claims are gone and the file is down to 3 comment lines.

The `destinationLabel` change was a good call beyond what I asked for — `ssw.com.au/products/rewards` reads as a real destination where the bare host read like a link to the homepage.

content/products/index/index.json
isaaclombardssw: Confirmed — back to `/products/rewards`. Soft navigation restored, no 301 hop, and preview slots keep reviewers on the preview.

tsconfig.json
isaaclombardssw: Confirmed — `tsconfig.json` is out of the diff entirely. `moduleResolution: node` and `jsx: preserve` match `main`.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — field removed from `tina/collections/products.tsx`, key removed from `index.json`, lock regenerated. That is the right end of it: no mandatory CMS field that renders nowhere.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed, and the sort still earns its keep even though `MoreProductsPanel` was deleted since. Brand-first is what stops a `col-span-2` card being bumped to the next row and leaving a hole in the one above — that outlives the panel it was originally for. Stable sort, CMS order otherwise preserved.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — one `BRAND_CARD_CONFIGS` map now drives the sort, the render branch and the per-brand config, with Josh's `satisfies` on top.

One leftover: `brandConfigFor` indexes that map with a plain `string`, so under `strict: false` the lookup comes back implicitly `any` and the literal-key inference `satisfies` bought gets thrown away at the only place it would have been used. Annotating the return `: BrandCardConfig | undefined` gets it back. Not worth a round on its own.

content/products/index/index.json
isaaclombardssw: Confirmed — "Quizzes". It is still the third tag so it never renders at `MAX_VISIBLE_TAGS = 2`, but the new Tina field description states that outright, which is the right fix for the editor-facing half.

__tests__/components/products/productCard.tsx
isaaclombardssw: Agreeing with Ken on this one. `__tests__` is a tracked pattern here and CI actually runs it (`pr-tests.yml` → `pnpm test-and-coverage`), so it has to be committed to do anything; `.gitignore` only excludes `/test-results/`, which is Jest's output.

The test is also a real regression guard, not scaffolding: `CustomLink` returns a bare fragment when `href` is falsy, so a product saved without a `url` would drop the element carrying the card chrome and spill its children into the grid as loose items. Worth keeping.

One nit for Ken while it is open: the file is 28 comment lines to ~115 of code, and 99-104 is revision narrative ("an earlier revision used the custom `text-xxs` key") — the same thing we cut from the components.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — 8 comment lines in 81 now, and the three that are left each guard a real gotcha (the dark-mode white band, the `h1` margin, `col-span-2` at the base grid tier).

public/images/company-logos/TinaCms-Logo-Full-Default.svg
isaaclombardssw: Agreeing with Ken here. It is the genuine TinaCMS lockup, not hand-authored — SVG is XML, so any vector asset lands in a diff as markup regardless of where it came from. Nothing to change.

styles.css
isaaclombardssw: Flagging a conflict between our two reviews so Ken is not stuck between them: I asked for the contrast values to stay, so the three that remain are mine, not leftovers. Happy to defer if you would rather they go.

Either way one of them is dead: `--brand-tina-orange` and the `brand-tina` Tailwind token it feeds are declared and never used — only `-field` and `-field-hover` are. Ken, delete `styles.css:131` and `tailwind.config.js:365` and a third of the comment goes with them.

public/images/YakShaver-Gradient-BG.png
isaaclombardssw: Not done, and I think that is fair — a hand-tuned crop and angle is a reasonable reason not to regenerate it.

It is 164 KB of PNG for a 500×320 gradient though, which is about the worst format choice for that image. A WebP export is a fraction of the bytes for pixel-identical output. Cheap win if you are in the file anyway.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — `tinaNode` is gone completely: zero references left, the 9 guarded ternaries with it, and `tinaField(product, ...)` is called directly. `product` is typed `ProductsIndexProductsList` at both card call sites.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — `ProductsIndexContent({ data }: { data: ProductsIndexQueryQuery })`, and `node` is `productsIndex` throughout. Field access is checked against the real schema now, so a collection rename fails at build.

Ken's follow-up note is right: `products-index-preview.tsx:6` still has the untyped `({ props })` one layer up. Agree that belongs in a separate PR rather than growing this one.

app/(about)/products/products-index.tsx
isaaclombardssw: Confirmed — applied as written. One caveat on the same map, noted on the `BRAND_CARD_CONFIGS` thread: `brandConfigFor` indexes it with a plain `string`, so the lookup widens straight back to `any`. That is the last mile of this suggestion.

components/products/yakShaverProductCard.tsx
isaaclombardssw: Confirmed — both files are gone, replaced by one `brandProductCard.tsx` driven by `TINA_CARD_CONFIG` / `YAKSHAVER_CARD_CONFIG`. Verifying the merge by diffing rendered HTML against the old components was the right way to do it.

The one wart left is `contentWrapperClassName?` — an optional wrapper that exists only because YakShaver does not group its content the way TinaCMS does. It is documented, and it is far better than two near-identical files.

components/products/tinaProductCard.tsx
isaaclombardssw: Confirmed — the paragraph exists once now, on `brandCardPadding` in `shared.ts:15-18`, and the constraint is encoded in the token instead of being trusted to a comment in two places.

components/products/productCard.tsx
isaaclombardssw: Confirmed — down to two types, both `{ product: ProductsIndexProductsList }`, derived from the generated type rather than restating it. Adding a field to `productsList` needs no hand edits now.

isaaclombardssw (commented):
## 🤖 Agent Review — follow-up round

Replied on every open thread, mine and Josh's. **All eight of my items and all six of Josh's code items are done** — verified against `b56aa1fe`, not taken on the reply. CI is green across the board.

The two structural ones are genuinely better than what I asked for: the two brand card components collapsed into one config-driven `brandProductCard.tsx`, and `tinaNode` is gone completely rather than just typed.

### 🤏 Still open
- **Dead token** — `--brand-tina-orange` (`styles.css:131`) and the `brand-tina` Tailwind entry (`tailwind.config.js:365`) are declared and never used. Two-line delete, and it takes a third of the comment Josh flagged with it.
- **PR description does not match the branch** — Josh's point from last round, still unfixed. It describes `MoreProductsPanel` and separate TinaCMS/YakShaver cards, none of which exist now, and `- Affected routes:` / `Fixed #{{ ISSUE NUMBER }}` are still placeholders. It also doesn't mention the new `tags` field, the `subTitle` removal, the Turbopack root pin in `next.config.mjs` or the Jest alias.

### 📞 Callouts
- `__tests__` and the SVG: I'm siding with Ken, both pushbacks are correct — details on those threads.
- The YakShaver gradient is 164 KB of PNG for a 500×320 gradient. WebP if you're in the file, not worth a round on its own.
- Comment volume landed at ~9% for `components/products/`. That's where I wanted it. The test file at 28/115 is the one place it didn't quite go.

components/products/productCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+            loading="lazy"
+            placeholder="blur"
+            blurDataURL={BluredBase64Image}
+            className="size-12 object-contain"
+          />
+        )}
+      </div>
+
+      <div className="flex min-w-0 flex-col gap-1">
+        <h3
```
isaaclombardssw: **Heading order skips a level, and this PR is what caused it.**

Production today runs `h1` (title) -> `h2` (the subtitle) -> `h3` (cards). Removing `subTitle` was the right call and I asked for it, but it takes the only `h2` off the page, so it now goes `h1` straight to these `h3`s. That is the axe `heading-order` rule ("Heading levels should only increase by one"), which Lighthouse runs as part of its accessibility audit.

`/consulting`, the page this one is mirroring, does not have the problem - it has real `h2` category headings between the `h1` and its card `h3`s (`app/consulting/index.tsx:252`).

There is no natural category level here, so the fix is just to make the card titles `h2`: this line and the `sr-only` heading at `brandProductCard.tsx:66`. Nothing visual changes - the size is set by `text-lg`, not the tag.

components/products/brandProductCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+        brandCardPadding
+      )}
+    >
+      <Image
+        src={config.watermark.src}
+        alt=""
+        aria-hidden
+        width={config.watermark.width}
+        height={config.watermark.height}
+        loading="lazy"
```
isaaclombardssw: **These are the LCP images, and they are explicitly lazy.**

The brand-first sort pins TinaCMS and YakShaver to the first row by construction, so both brand cards are above the fold at every breakpoint - and `YakShaver-Gradient-BG.png` is a 164 KB full-height PNG. `loading="lazy"` on the LCP element is its own Lighthouse audit ("Largest Contentful Paint image was lazily loaded"): the browser will not preload it, so it is fetched after layout instead of during it.

Drop the `loading="lazy"` here and add `priority` to the watermark and the lockup on both brand configs. The standard `ProductCard` logos below can stay lazy - those genuinely are below the fold.

Separately on the same asset: 164 KB of PNG for a 500x320 gradient is the worst-case format choice. A WebP export is a fraction of the bytes for pixel-identical output, if you are in the file anyway.

isaaclombardssw (commented):
## 🤖 Agent Review — re-review of `b56aa1fe`

**🟡 Approve with minor changes.** Everything from my last two rounds and Josh's code-quality round is done and verified against the branch, not taken on the reply. CI is green and I ran the suite locally from a clean `--frozen-lockfile` install (4 suites, 35 tests).

Three things left to update, two commented inline.

### 🚨 Must Fix / Address
- ✌️ **None** - nothing here blocks merge.

### ⚠️ Notable Changes
- ♿ **Heading order now skips h1 to h3** - commented inline on `productCard.tsx`. Deleting the `subTitle` h2 left the page with no h2 at all, which axe and Lighthouse flag, and `/consulting` (the page this mirrors) keeps one.
- **Above-the-fold brand artwork is lazy-loaded** - commented inline on `brandProductCard.tsx`. The sort pins both brand cards to the first row, so these are the LCP images.
- **The description still does not describe this branch** - Josh raised this last round and it is unchanged. It advertises a `MoreProductsPanel` that no longer exists, does not mention the `subTitle` schema deletion, the new `tags` field, the copy rewrites, the Turbopack root pin or the Jest alias, and `- Affected routes:` / `Fixed #{{ ISSUE NUMBER }}` are still placeholders. Worth knowing that `Check linked issues` passed by accident: the head is a merge commit whose first-parent diff was content-only, so the gate skipped itself and there is genuinely no linked issue.

### 📞 Callouts
- **The brand-first sort is not dead code** - a reviewer suggested deleting it as a no-op. I simulated grid auto-placement: without it, one reorder in the CMS leaves a visible skipped cell mid-grid at both the 2-column and 4-column tiers; with it, none. Please keep it.
- **The `p-6 pr-10` token is worth keeping too** - same reason it was introduced. `twMerge("pr-10 p-6")` collapses to `p-6`, silently dropping the right padding. Verified by running it.
- **A rename does not break the grid** - if a product name stops matching, it falls through to a standard card. Graceful, so I have not raised the name-keying as something to fix.

### 🤏 Small Feedback
- 🎨 **Dead token** - `--brand-tina-orange` and the `brand-tina` Tailwind key are defined and never consumed; only `-field` and `-field-hover` are used. Two-line delete.
- **`PROD_BASE_URL` already exists** in `components/util/constants.tsx:10`, where `destinationLabel` hardcodes `https://www.ssw.com.au`.
- **A product named `constructor` would crash the page** - the brand map is a plain object literal, so prototype members resolve truthy and then throw on `config.lockup.src`. `Object.hasOwn` closes it. Unlikely, one line.
- 14 more optional nits (surface classes duplicated from `consultingCard`, `ProductCardShell` being the sixth hand-rolled link-or-div, test file comment volume). Happy to post the list if useful - none of them need to hold up merge.

hveraus:
> ## 🤖 Agent Review — follow-up round
> Replied on every open thread, mine and Josh's. **All eight of my items and all six of Josh's code items are done** — verified against `b56aa1fe`, not taken on the reply. CI is green across the board.
> 
> The two structural ones are genuinely better than what I asked for: the two brand card components collapsed into one config-driven `brandProductCard.tsx`, and `tinaNode` is gone completely rather than just typed.
> 
> ### 🤏 Still open
> * **Dead token** — `--brand-tina-orange` (`styles.css:131`) and the `brand-tina` Tailwind entry (`tailwind.config.js:365`) are declared and never used. Two-line delete, and it takes a third of the comment Josh flagged with it.
> * **PR description does not match the branch** — Josh's point from last round, still unfixed. It describes `MoreProductsPanel` and separate TinaCMS/YakShaver cards, none of which exist now, and `- Affected routes:` / `Fixed #{{ ISSUE NUMBER }}` are still placeholders. It also doesn't mention the new `tags` field, the `subTitle` removal, the Turbopack root pin in `next.config.mjs` or the Jest alias.
> 
> ### 📞 Callouts
> * `__tests__` and the SVG: I'm siding with Ken, both pushbacks are correct — details on those threads.
> * The YakShaver gradient is 164 KB of PNG for a 500×320 gradient. WebP if you're in the file, not worth a round on its own.
> * Comment volume landed at ~9% for `components/products/`. That's where I wanted it. The test file at 28/115 is the one place it didn't quite go.

✅ Dead token — done
Verified unused before deleting rather than taking the review's word:

--brand-tina-orange  → only referenced by tailwind's "brand-tina" entry
brand-tina           → zero usages anywhere
brand-tina-field(-hover) → used, brandProductCard.tsx:132
So the two dead declarations went, and the field colours stayed. The comments they inflated came down with them — and the styles.css one had a second problem the review didn't mention: it said "Third-party brand colours (TinaCMS, YakShaver)", but no YakShaver token was ever declared. It's now one line naming only TinaCMS.

hveraus:
> ## 🤖 Agent Review — re-review of `b56aa1fe`
> **🟡 Approve with minor changes.** Everything from my last two rounds and Josh's code-quality round is done and verified against the branch, not taken on the reply. CI is green and I ran the suite locally from a clean `--frozen-lockfile` install (4 suites, 35 tests).
> 
> Three things left to update, two commented inline.
> 
> ### 🚨 Must Fix / Address
> * ✌️ **None** - nothing here blocks merge.
> 
> ### ⚠️ Notable Changes
> * ♿ **Heading order now skips h1 to h3** - commented inline on `productCard.tsx`. Deleting the `subTitle` h2 left the page with no h2 at all, which axe and Lighthouse flag, and `/consulting` (the page this mirrors) keeps one.
> * **Above-the-fold brand artwork is lazy-loaded** - commented inline on `brandProductCard.tsx`. The sort pins both brand cards to the first row, so these are the LCP images.
> * **The description still does not describe this branch** - Josh raised this last round and it is unchanged. It advertises a `MoreProductsPanel` that no longer exists, does not mention the `subTitle` schema deletion, the new `tags` field, the copy rewrites, the Turbopack root pin or the Jest alias, and `- Affected routes:` / `Fixed #{{ ISSUE NUMBER }}` are still placeholders. Worth knowing that `Check linked issues` passed by accident: the head is a merge commit whose first-parent diff was content-only, so the gate skipped itself and there is genuinely no linked issue.
> 
> ### 📞 Callouts
> * **The brand-first sort is not dead code** - a reviewer suggested deleting it as a no-op. I simulated grid auto-placement: without it, one reorder in the CMS leaves a visible skipped cell mid-grid at both the 2-column and 4-column tiers; with it, none. Please keep it.
> * **The `p-6 pr-10` token is worth keeping too** - same reason it was introduced. `twMerge("pr-10 p-6")` collapses to `p-6`, silently dropping the right padding. Verified by running it.
> * **A rename does not break the grid** - if a product name stops matching, it falls through to a standard card. Graceful, so I have not raised the name-keying as something to fix.
> 
> ### 🤏 Small Feedback
> * 🎨 **Dead token** - `--brand-tina-orange` and the `brand-tina` Tailwind key are defined and never consumed; only `-field` and `-field-hover` are used. Two-line delete.
> * **`PROD_BASE_URL` already exists** in `components/util/constants.tsx:10`, where `destinationLabel` hardcodes `https://www.ssw.com.au`.
> * **A product named `constructor` would crash the page** - the brand map is a plain object literal, so prototype members resolve truthy and then throw on `config.lockup.src`. `Object.hasOwn` closes it. Unlikely, one line.
> * 14 more optional nits (surface classes duplicated from `consultingCard`, `ProductCardShell` being the sixth hand-rolled link-or-div, test file comment volume). Happy to post the list if useful - none of them need to hold up merge.

Fixed
1. Heading order skipped h1 → h3 — [productCard.tsx:53](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/components/products/productCard.tsx#L53), [brandProductCard.tsx:66](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/components/products/brandProductCard.tsx#L66)

Confirmed: deleting subTitle left the page with no h2 at all. Rather than adding an invisible h2, I promoted the card titles from h3 to h2. On /consulting the chain is h1 → h2 (category) → h3 (card); /products has no category grouping, so the cards sit directly under the page h1 and h2 is the honest level. The test uses getByRole("heading", { name }) with no level assertion, so it was unaffected.

2. Above-the-fold brand artwork was lazy-loaded — [brandProductCard.tsx:105-116](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/components/products/brandProductCard.tsx#L105-L116)

Swapped the watermark's loading="lazy" for priority, and added priority to the lockup too — it had no loading attribute, but next/image defaults to lazy, so it had the same problem. The sort pins both brand cards to the first row, making these the LCP candidates.

3. Hardcoded base URL — [shared.ts:52](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/components/products/shared.ts#L52)

Now imports PROD_BASE_URL from [constants.tsx:10](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/components/util/constants.tsx#L10).

4. A product named constructor would crash the page — [products-index.tsx:20-28](vscode-webview://1laj2o3926bl1aetdep7acq69e44315t6r1qaboc6e3uruopmu1d/app/(about)/products/products-index.tsx#L20-L28)

Closed with Object.hasOwn. The browserslist floor is Chrome 111 / Safari 16.4, well above Object.hasOwn support (Chrome 93 / Safari 15.4).

components/products/productCard.tsx
hveraus: Fixed in 7dd19dab1. Went with your suggestion — card titles are now `h2` in both `productCard.tsx` and the `sr-only` heading in `brandProductCard.tsx`, so the outline is h1 → h2 with nothing in between.

Confirmed there's no other heading source on the page: `Breadcrumbs`, `HomeThemeShell` and `Container` emit none, so those two files were the whole fix. And as you said, nothing moves visually — the size is `text-lg`.

components/products/brandProductCard.tsx
hveraus: Fixed in b6a9be445. `loading="lazy"` is gone and both the watermark and the lockup now use `priority`. Since both brand cards share the one component, that covers TinaCMS and YakShaver together. Standard `ProductCard` logos left lazy, as you suggested.

On the PNG — converted, but flagging that the byte figure doesn't reflect what ships. Image optimisation is on in `next.config.mjs` (no `unoptimized`, no `formats` override), so next/image was already re-encoding this through `/_next/image` as WebP at roughly 23 KB. Users were never downloading 164 KB.

Converted anyway since it's a real repo and Docker image saving: WebP q82, 164,023 → 24,844 bytes, alpha intact. Checked for gradient banding — only 5.7% of the image is opaque, and across those pixels the worst channel deviation from the source is 4/255 (mean 0.79), so the extra generation loss from the optimiser re-encoding it is not visible.

One caveat: the PNG blob still exists in this branch's intermediate commits, so the saving is only complete if this merges as a squash.

components/products/brandProductCard.tsx
```diff
@@ -0,0 +1,179 @@
...
+        src={config.lockup.src}
+        alt=""
+        aria-hidden
+        width={config.lockup.width}
+        height={config.lockup.height}
+        priority
+        className={config.lockup.className}
+      />
+      {/* The lockup is an image, so this carries the name for a11y/SEO.
+          h2 to match ProductCard — the cards sit directly under the page h1. */}
```
isaaclombardssw: Last bit of cleanup, we can really cut down on comments still! It only makes it harder to read and should be reserved for genuine non-standard code, tech debt or complex workarounds

isaaclombardssw (changes requested):

components/products/brandProductCard.tsx
hveraus: Done in 77b3eef36 — took the whole set rather than just this one. 46 comment lines across `components/products/` and `products-index.tsx` down to 26, applying your bar of non-standard code, tech debt or complex workarounds.

Cut: the file and component headers that restated the exports, the `BrandCardConfig` field docs that restated the field names, the `visibleTags` pipeline description sitting directly above the pipeline, the `ArrowCircle` fill note, and both heading-level comments I added in 7dd19dab1 — you're right, an `h2` card title is ordinary markup, not a workaround. `productCard.tsx` is comment-free now.

Kept the ones where the code looks wrong until you know why, and where deleting the comment invites someone to "fix" it back:

- `border-0.75` — `borderWidth.DEFAULT` is 3px in this repo
- the twMerge `outline` and `p-6 pr-10` notes (the second is the one you asked to keep)
- `CustomLink` returning a bare fragment on a falsy `href` — the reason `productCardShell.tsx` exists
- the literal `dark` class on a brand surface
- span-not-button, since both are invalid nested in an `<a>`
- `Object.hasOwn`, which otherwise reads like a pointless wrapper
- `min-h-screen`, `m-0 p-0` and `col-span-2` — each one a global-CSS or grid trap

isaaclombardssw (approved):
✅ Good work with this :)

PR #4987, opened 2026-08-13
fix(ui): one chevron breadcrumb separator, defined by the primitive

### LINKED ISSUE #4988: Make sure the primitive breadcrumb is following the design system
Claude reported a slash instead of a chevron

tiagov8:
Closes #4988

Follow-up to #4978.

[design.ssw.com.au](https://design.ssw.com.au/?path=/docs/ui-atoms-breadcrumb--docs) specifies a **chevron** between crumbs. The site had three different separators, none of them a chevron.

## Changes
- `components/ui/breadcrumb.tsx` — the primitive now defaults to `ChevronRight` (was a `Slash` icon). **This is the single source of truth.**
- `app/components/breadcrumb.tsx` — dropped its `"/"` override, so it inherits the default.
- `components/blocks/breadcrumbs/breadcrumbs.tsx` — dropped its hand-drawn slash SVG override and deleted the now-dead `Separator` component (plus an unused `cn` import).

No call site overrides the separator any more, so changing it in one place changes it everywhere.

## Impact
Every breadcrumb on the site — consulting, products, articles, offices, industry, company, employment, training, netug, events — now shows a chevron.

## Note
`app/(events)/events/[...filename]/events-preview.tsx` still has a `">"` override on `main`. It isn't touched here to avoid conflicting with #4978, which deletes that hand-rolled breadcrumb outright in favour of the shared component.


isaaclombardssw (approved):
LGTM

PR #4992, opened 2026-08-20
Advertise .NET 10 as the LTS

isaaclombardssw:
`/events/net-superpowers-tour` said ".NET 9 is the latest Long-Term Support (LTS) version in .NET". That was never true — .NET 9 is an **STS** release and hits end of support on 10 Nov 2026. .NET 10 (Nov 2025) is the current LTS, supported until Nov 2028.

Two other pages had the same problem, one of them worse. Everything here is corrected against Microsoft's own release notes (linked below).

### .NET Superpowers Tour
- .NET 9 → .NET 10, C# 13 → C# 14, LTS sentence now names .NET 10 and its Nov 2028 support date
- **".NET Aspire" → "Aspire"** — the product dropped the .NET prefix and is on 13.x; it orchestrates Python and JS services now, so it's no longer .NET-only
- Agenda refreshed, because it was .NET 8/9-era content sitting under bumped version numbers:
  - *Minimal APIs* — "the new ASP.NET Core Identity" meant .NET 8's Identity API endpoints. .NET 10's Identity story is **passkeys/WebAuthn**; also added **built-in Minimal API validation** and **OpenAPI 3.1 from XML comments**
  - *AI* — `Microsoft.Extensions.AI` isn't "new" anymore, so name it, and add the **Microsoft Agent Framework**
  - *Testing* — replaced the contentless "What's new in testing" bullet with **Microsoft.Testing.Platform**, now natively supported by `dotnet test` (opt-in via `global.json`)
- Typo fixes: NSubtitute, Whats

### .NET App Modernization Assessment
Still pitching .NET 8 as "the latest LTS". All .NET 8 → .NET 10, C# 12 → C# 14, and the rule link repointed to `rules-to-better-net10-migrations` (verified it exists).

### .NET Upgrade FAQ
The "Is migrating to .NET 10 worthwhile?" answer claimed ".NET 10 is a Current release (not Long-Term Support)" — the opposite of the truth, and it contradicted another answer in the same accordion. Also corrected STS support length from 18 to 24 months in two places; Microsoft extended it.

### Still outstanding
- **Dates.** The tour page still shows July 2025 with dead Eventbrite links. There's no 2026 .NET Superpowers event in `content/events-calendar/`, so there was nothing real to point them at. Needs dates + booking URLs.
- **MCP is listed twice** in the agenda — a Stage 5 bullet and all of Stage 8. Predates this PR.
[...]

PR #4993, opened 2026-08-20
💄 Update .NET Superpowers hero logo to .NET 10

isaaclombardssw:
#4992 moved `/events/net-superpowers-tour` to .NET 10, but the hero overlay still read ".NET 9".

Only `DevSuperPowers-DOTNET.png` carried a version number — the OG image is just ".NET" with no digit, and the alt text is version-agnostic, so neither needed changing.

### What changed

Erased the "9" and its 20%-black halo, rendered "10" at the same cap height (234px), baseline, left edge and halo radius measured off the old glyph, then re-centred the arc over the wordmark and the whole lockup on the canvas.

Renamed to `DevSuperPowers-DOTNET10.png`. Keeping the old path with new bytes means Tina, the CDN and `next/image` can all keep serving the stale .NET 9 asset — which is what happened last time, when the .NET 8 → 9 update overwrote the same filename.

### Caveat

The original digit's typeface isn't in [dotnet/brand](https://github.com/dotnet/brand) and isn't installed locally — the closest match was Figtree Black, at 94% glyph overlap against the old "9". At the 680px this renders at in `eventsHeader.tsx` it's indistinguishable, but a designer redoing it from the source file would be the proper fix.

### Still outstanding from #4992

- Dates still show July 2025 with dead Eventbrite links; no 2026 event exists in `content/events-calendar/`.
- MCP is listed twice in the agenda.


0xharkirat (approved):
LGTM

PR #5008, opened 2026-09-02
💄 Redesign /events index - with consulting sidebar, product card chrome, dark mode

hveraus:
Brings `/events` in line with the pages either side of it. It had no dark mode while `/` and `/consulting` did, and its cards were a floated 100px thumbnail over a bare bottom border — the last index page still on the old chrome.

- Affected routes: `/events`, `/consulting`

There is no linked issue for this one.

## What changed

**Layout** — borrowed from the `/consulting` index
- New [components/events/eventsSidebar.tsx](components/events/eventsSidebar.tsx): the sticky index, with the page `h1` in the sidebar, `grid-cols-sidebar`, and the mobile scrim strip that cards scroll under. The existing filter groups (Technology / Format, with counts) render as its nav.
- Deliberately **not** folded into `FilterBlock`: that is shared with `/clients` and `/opportunities`, which are unthemed and want the old flex column. Variant flags there would have made it worse for both.
- `sidebarBody` (the ~800px Video On Demand promo) moved out of the sidebar to the end of the content column. Inside the sticky column it made the sidebar 1343px tall in a 900px viewport, so scrolling dragged it straight over the pinned nav.

**Cards** — the home Upcoming Events block, wearing the `ProductCard` chrome
- `cardShell` and `productTagChip` are imported from `components/products/shared.ts` rather than re-approximated. The local `eventTagChip` this replaced was the same class string character for character.
- Same surface, hover and active states as `ProductCard`, so hover raises the brand border here too. Same `rounded-card` plate, same `ArrowCircle` override (`size-9`, gray fill, `size-3.5` icon) at bottom-right.
- Icon meta row (date, location, presenter, tags) replaces the old inline `Presenter:` / `Location:` labels.

**Thumbnails now have a standard ratio.** 29 of the 40 files in `public/images/events` are 1:1; the rest scatter from 1.07 to 3.69. The previous full-height `object-cover` cropped the square majority and let the image stretch to whatever the content height was. Now a fixed 1:1 plate with `object-contain`, so every ratio shows whole, on a white plate for the transparent logos among them.

**Theming** — `/events` added to `THEMED_ROUTES`, so it gets the pre-paint script and the same light/dark treatment as `/` and `/consulting`.

**Sidebar row spacing** tightened on `/events` *and* `/consulting` together, from a 50px pitch to 38px, so the two stay in lockstep. Mobile touch targets are unchanged — the pill row's `py-2.5` already renders ~40px, so the smaller `min-height` never binds there.

## Styles

`styles.css` is untouched. **`tailwind.config.js` is not** — an earlier revision of this description claimed otherwise, which was wrong:

- The shared `sidebar` / `sidebar-narrow` grid tokens are retuned, from `minmax(280px, 30%)` to `minmax(300px, 26%)` and `minmax(240px, 30%)` to `minmax(260px, 30%)`. **This resizes `/consulting` too**, since it reads the same tokens.
- Two tokens are new: `min-h-sidebar-card` (200px) and `max-w-sidebar-card` (300px), for the promo card.

Everything else was already there: `top-headerOffset`, `sunken-glow`, `sunken-scrim`, `rounded-card`, `hairline`, `card`/`card-hover`, `brand-subtle`.

**Sidebar rows are shorter on desktop as well as tighter.** The extracted `sidebarNavItem` is `min-h-9` where `/consulting` had `min-h-11`, so desktop rows go 44px → 36px. Mobile is unaffected — the pill row's `py-2.5` already renders ~40px, so the min-height never binds there.

## Measured, at 1440x900

| | before | after |
|---|---|---|
| card height (avg) | 277px | 173px |
| page scroll height | 4022px | 3161px |
| meta block | 93px | 54px |
| sticky sidebar | 1343px | 504px |

A 4-column meta grid was tried first and was worse (107px) — narrow cells wrap more than the row saves. `flex-wrap` won.

Overlap fix verified at scroll 2200 / 3000 / 3800: six sample points down the pinned nav all resolve to the nav itself, where before they returned the promo's ``.

## Deliberate removals

Called out so they don't read as accidents:

- **The Video On Demand / Vimeo promo content is gone**, and with it the five `*OnDemand` / `*OnVimeo` images it was the only reference to (~180KB, deleted in this PR).
- **Presenter and office links are gone from the cards.** The whole card is one anchor, so a nested profile link would be unreachable by pointer while still sitting in the tab order. `PresenterList` takes a `linkless` prop for this, and there is now a test pinning the anchor count with and without it.
- **`sidebarBody` is deleted from the schema**, not just relabelled. Its content was already empty and `afterEvents` covers the same slot, so it was a second editor field pointing at the same place.

## Worth a reviewer's opinion

**The card no longer shows the event description.** `line-clamp-2` does not clamp `TinaMarkdown`'s block children, so it rendered inconsistently between 54px and 67px, and it was the largest remaining block on an index card. Happy to put it back as a first-paragraph excerpt if you would rather keep it.

**`components/filter/events.tsx` now imports from `components/products/shared.ts`.** The module is pure constants and helpers — no React, no server imports — so it is safe, but the `products` namespace on something `/events` consumes is a bit off. Happy to promote `cardShell` / `productTagChip` to a neutral module if preferred.

## Screenshots

`1835b01` against its base `b20789d`, so the event list is identical on both sides and only the chrome differs.

### Desktop — 1440x900

**Before**

The old /events index: a plain filter column with Video On Demand under it, and rows of events separated by a bottom border, each with a floated thumbnail, bold Presenter / Location / Type / Category labels, a description paragraph and a grey Find out more button

**After — light**

The redesigned /events index: a sticky SSW Events sidebar carrying the h1 and the Technology and Format filters, and cards on the ProductCard shell with a square thumbnail plate, an icon meta row and a circled arrow at bottom right

**After — dark**, which `/events` had no version of before

The same redesigned /events index in dark mode: near-black page, dark card surfaces with hairline borders, and the mega menu switched to its dark treatment

**One card, up close** — the `ProductCard` chrome, the 1:1 `object-contain` plate, and the icon meta row that replaced the `Presenter:` / `Location:` labels.

A single event card: square .NET User Group logo on a white plate, title, then a calendar icon with the date and a 6 DAYS TO GO badge, a person icon with the presenter, a pin icon with the venue, and a tag icon with two category chips

### Mobile — iPhone 14 Pro, 393x660

**Before.** The filter column is a full-height stack with the Video On Demand promo under it, so the opening viewport is all chrome and the first event is roughly 1400px down.

The old /events index on a phone: the SSW Events heading, an I am looking for... label and the All Technology filter list running off the bottom of the screen, with no event visible

The same page scrolled 1400px down: the tail of the Video On Demand promo, the Upcoming / Past Events tabs, and only then the first event, with its Presenter / Location / Type / Category labels stacked one per line

**After — light.** The two filter groups collapse into horizontally scrolling pill rows, and the first card is in the opening viewport.

The redesigned /events index on a phone: SSW Events with All Technology and All Formats pill rows beneath it, then Upcoming Events and the first card already on screen

**After — dark**

The same phone view in dark mode: black page, dark pills and a dark card surface

**After, scrolled to the same 1400px.** The pill rows pin to the top and cards pass under the scrim strip — at this offset you are several cards down the list, where `main` was still arriving at the first.

The redesigned index scrolled down: the two pill rows pinned at the top with a card title fading out underneath them, and the Newcastle Coders card fully visible below

- [x] Include Done Video or screenshots


components/filter/events.tsx
```diff
@@ -330,84 +436,100 @@ const Event = ({ visible, event, jsonLd }: EventProps) => {
...
+            "dark:border-hairline dark:hover:border-brand",
+            "active:bg-gray-100 dark:active:bg-card"
+          )}
+        >
+          {/* The whole card is the link. Everything else stays below it, so
+              nested interactive elements can't end up inside an <a>. */}
+          <CustomLink
+            href={event.url}
+            aria-label={`Find out more about ${event.title}`}
+            className="unstyled absolute inset-0 z-10 rounded-card !no-underline focus-visible:outline focus-visible:outline-offset-2 focus-visible:outline-brand"
```
isaaclombardssw: 🚨 **The focus ring never paints on any event card.**

This anchor is `absolute inset-0` inside `cardShell`, which sets `overflow-hidden`. `outline-offset-2` puts the ring 2px outside the anchor's box, which is exactly the wrapper's padding box, so the wrapper clips all of it.

Verified on a running dev server, not just a reduced case. The card focuses, `:focus-visible` matches, `outline-style: solid` is applied, and nothing is drawn. Setting `overflow: visible` on that one wrapper in devtools makes the ring appear immediately.

`ProductCard` dodges this only because there the anchor **is** the card — an element's own `overflow` never clips its own outline. That's the fix here too: use `ProductCardShell` with `href={event.url}` and drop the overlay and the `z-10`/`z-20` layering. `linkless` already removed every nested anchor, so the overlay has nothing left to justify it, and you get real anchor text and selectable card text back at the same time.

Side note: even unclipped the ring is `outline-width: 1px`, which is very faint for a focus indicator.

components/filter/events.tsx
```diff
@@ -330,84 +436,100 @@ const Event = ({ visible, event, jsonLd }: EventProps) => {
...
+            // modes; only the border colour still differs.
+            "border-stroke-weak bg-card hover:border-brand hover:bg-card-hover",
+            "dark:border-hairline dark:hover:border-brand",
+            "active:bg-gray-100 dark:active:bg-card"
+          )}
+        >
+          {/* The whole card is the link. Everything else stays below it, so
+              nested interactive elements can't end up inside an <a>. */}
+          <CustomLink
+            href={event.url}
```
isaaclombardssw: An event saved without a URL renders a card that looks clickable and isn't.

`CustomLink` returns a bare fragment when `href` is falsy, and this one has no children, so nothing at all renders. The wrapper keeps `hover:border-brand` and `hover:bg-card-hover`, so the card still lights up on hover with nothing behind it.

Not hypothetical — `getEvents.ts:84` guards `event.url && fixRelativeUrl(event.url)`, so the data can be falsy. There's already a regression test for this exact `CustomLink` trap on `ProductCard` (`__tests__/components/products/productCard.tsx`). Moving to `ProductCardShell` fixes this one too, since it falls back to a plain `div`.

components/filter/events.tsx
```diff
@@ -279,6 +336,47 @@ const LoadedEvents: React.FC<AllEventsProps> = ({
   );
 };
 
+const EventMetaItem = ({
```
isaaclombardssw: `EventMetaItem`, `EventMetaGrid` and the days-to-go badge on line 491 are re-implementations of the unexported ones in `components/blocks/v3/events/events.tsx` — the badge class string is identical character for character.

Worth lifting all three into something like `components/events/eventMeta.tsx` and importing in both, before they drift. They already differ slightly: this badge is always red, while `EventsRelativeBox` renders it grey unless the event is running or today, so the same badge looks different here and on the home page.

components/presenters/presenterList.tsx
```diff
@@ -7,9 +7,17 @@ export type Presenter = {
 
 type PresenterListProps = {
   presenters: { presenter?: Presenter }[];
+  // Renders names as plain text instead of profile links. For callers that sit
+  // inside a larger click target: a nested anchor under an overlay link is
+  // unreachable by mouse but still focusable, so the two inputs would navigate
+  // to different places.
+  linkless?: boolean;
```
isaaclombardssw, line 10: `linkless` is what resolves the accessibility regression Copilot flagged, and it's one render assertion to pin: render with and without, count the anchors. Cheap insurance against someone removing the prop later and quietly reintroducing unreachable-but-focusable links.

components/filter/FilterNav.tsx
```diff
@@ -0,0 +1,85 @@
...
+type FilterNavProps = FilterGroupProps & { label: string };
+
+const FilterNav = ({
+  label,
+  allText,
+  options,
+  selected,
+  setSelected,
+}: FilterNavProps) => (
+  <nav aria-label={label} className="not-first:mt-5 max-md:not-first:mt-2">
```
isaaclombardssw, line 23: `label` is `group.allText` (line 80), so these landmarks come out as "All Technology navigation" and "All Formats navigation". Passing a separate `label` ("Technology", "Format") from `hooks/useEvents.ts` would read better.

Minor, same area: `aria-current="location"` on buttons that filter a list rather than navigate. `/consulting` earns that markup because those are real anchors to sections; here `aria-pressed` is closer to what's actually happening.

tailwind.config.js
```diff
@@ -76,9 +76,9 @@ export default {
         12: "repeat(12, minmax(min-content, 0fr))",
       },
       gridTemplateColumns: {
-        // Sticky-sidebar + fluid-content layout (used by /consulting).
-        sidebar: "minmax(280px, 30%) minmax(0, 1fr)",
-        "sidebar-narrow": "minmax(240px, 30%) minmax(0, 1fr)",
+        // Sticky-sidebar + fluid-content layout (/consulting and /events).
+        sidebar: "minmax(300px, 26%) minmax(0, 1fr)",
```
isaaclombardssw, line 80: The description says "Nothing added to `styles.css` or `tailwind.config.js` — the diff touches neither", but this retunes the shared `sidebar` token from `minmax(280px, 30%)` to `minmax(300px, 26%)`, which resizes `/consulting` too.

Not objecting to the change, just worth calling out in the description so nobody reviews this PR without realising `/consulting` moved. Same for the `min-h-11` → `min-h-9` in the extracted nav item: your note about mobile is right (`py-2.5` already renders ~40px so the min-height never binds), but desktop rows do go 44px → 36px.

isaaclombardssw (changes requested):
## 🤖 Agent Review

Nice consolidation — the `StickySidebarLayout` extraction is real (`/consulting` lost ~90 lines and gained nothing bespoke), the card chrome is reused rather than re-approximated, and every added Tailwind class resolves against the config with no arbitrary values. Dark mode on `/events` is overdue.

One blocking issue below: the card focus ring is invisible. I verified it on a running `pnpm dev`, not just in isolation — same card, same focus, ring appears only when `overflow-hidden` comes off the wrapper.

Confirmed with Isaac that the Vimeo content removal and the presenter/office link removal are **intentional**, so they're not counted against this. They should be stated in the description though, otherwise the next reader takes them for accidents.

tina/collections/events.tsx
```diff
@@ -122,7 +122,12 @@ export const eventsIndexSchema: Collection = {
     {
       type: "rich-text",
       name: "sidebarBody",
-      label: "Sidebar Body",
+      // Renders at the foot of the content column, below the events list, not
+      // in the sidebar. The field name is kept for the generated query and the
+      // existing content; only the editor-facing label describes the position.
+      label: "Body (below the events list)",
```
isaaclombardssw: Rather than relabel it, consider deleting `sidebarBody` outright.

The content is already empty and `afterEvents` right below it already fills this exact slot, so two editor fields now point at the same position. Removing it takes the schema field, the prop through `index.tsx` and `events.tsx`, the type on `EventsFilterProps`, and the render block — six code sites — and the lockfile is already regenerated in this PR anyway.

components/filter/events.tsx
```diff
@@ -102,63 +100,122 @@ export const EventsFilter = ({
...
+        <h2 className="m-0 p-0 text-xl font-semibold text-foreground max-md:text-lg xl:text-2xl">
+          {pastSelected ? "Past Events" : "Upcoming Events"}
+        </h2>
+        <TimeframeToggle
+          pastSelected={pastSelected}
+          onToggle={() => selectPast(!pastSelected)}
+        />
+      </div>
+
+      {pastSelected ? (
```
isaaclombardssw: These two arms are structurally identical, ~29 lines for one boolean.

Pick the five values once with a ternary (`events`, `isLoading`, `hasMore`, `fetchNext`, `isFetching`) and render a single `EventsList` + `LoadMore`. Worth adding `key={pastSelected ? "past" : "upcoming"}` so you keep the remount the old `TabPanel`s gave you for free.

components/filter/events.tsx
```diff
@@ -279,6 +336,47 @@ const LoadedEvents: React.FC<AllEventsProps> = ({
...
+      <span className="flex min-w-0 flex-wrap items-center gap-x-2">
+        {children}
+      </span>
+    </span>
+  );
+};
+
+// A column, not a wrapping row: each fact gets its own line, so the card
+// reads the same whether a title runs to one line or three.
+const EventMetaGrid = ({ children }: { children: React.ReactNode }) => (
```
isaaclombardssw: `EventMetaGrid` and `EventMetaRow` each wrap a single `div` and are each used once. Inlining them reads fine and drops two indirections.

(If you go the other way and share the meta primitives with the v3 block as suggested above, ignore this — sharing them is the better outcome.)

components/events/eventsSidebar.tsx
```diff
@@ -0,0 +1,85 @@
...
+  // Mounted in the sidebar on desktop and at the foot of the content column on
+  // mobile, both by StickySidebarLayout.
+  promo?: React.ReactNode;
+  children: React.ReactNode;
+};
+
+// The /events counterpart to /consulting's sticky category index. It is not
+// FilterBlock: that one is shared with /clients and /opportunities, which are
+// unthemed and lay the sidebar out as a plain flex column.
+export const EventsSidebar = ({
```
isaaclombardssw: `EventsSidebar` is 16 lines that forward straight to `StickySidebarLayout`.

`FilterNav` above it is typed purely on `FilterGroupProps` with nothing event-specific, and `components/filter/events.tsx` imports it back across the boundary. Moving `FilterNav` to `components/filter/FilterNav.tsx` as the themed sibling of `FilterGroup` and calling `StickySidebarLayout` directly from `events.tsx` deletes this wrapper and puts the generic bit where `/clients` and `/opportunities` can pick it up when they get themed.

app/(events)/events/index.tsx
```diff
@@ -13,34 +12,47 @@ import { TinaMarkdown } from "tinacms/dist/rich-text";
...
-          <h1 className="pt-0 md:mr-12 md:shrink-0 md:basis-64">SSW Events</h1>
-          <div className="mt-5 min-w-0 max-w-full shrink grow overflow-auto whitespace-normal break-all pb-1 pt-5 md:mr-12 md:shrink-0 md:basis-64">
-            <TinaMarkdown
-              content={data.eventsIndex.preface}
-              components={componentRenderer}
+      {/* min-h-screen, not min-h-full: PageLayout's <main> has an unconditional
+          bg-white, so any shortfall shows as a white band under the themed
+          content. */}
+      <HomeThemeShell className="min-h-screen bg-sunken-glow">
+        <div className="mx-auto max-w-8xl px-6 pb-16 pt-4 max-md:px-3 max-md:pb-12 max-md:pt-3">
```
isaaclombardssw: This wrapper is repeated verbatim on `/consulting`, which is the other consumer of the layout you just extracted. Folding it into `StickySidebarLayout` means the next padding tweak doesn't have to be applied twice by hand — which is the reason given for extracting the layout in the first place.

Same theme: `EventTrimmed` still carries `description` and `presenterProfileUrl` that nothing on this page renders now. Leave the query fragment alone since it's shared, but the type is wider than its only consumer.

content/events/index/index.mdx
```diff
@@ -5,40 +5,5 @@ seo:
     Learn from software industry professionals about the latest tech and best
     practices for MAUI, .Net, GitHub, DevOps.
   showBreadcrumb: true
-sidebarBody: >
```
isaaclombardssw, line 8: Five images are orphaned by this deletion — `AzureOnDemand`, `AngularOnDemand`, `CAOnDemand`, `dotNet5OnDemand` and `AzureOnVimeo` in `public/images/events`, about 180KB. Zero references left across `app/`, `components/` and `content/`. Worth removing in this PR while it's obvious why they're unused.

isaaclombardssw (commented):
## 🤖 Agent Review — optional simplifications

None of these block. Grouping them separately because they're all net deletions, so they're cheap and they make the next change to this file easier rather than harder.

The blocking fix is also a deletion, for what it's worth: moving to `ProductCardShell` removes the overlay anchor, the `z-10`/`z-20` layering and the `pointer-events-none` on the arrow, and fixes the dead-card case for free.

One more that's already noted on `EventMetaItem` above: reusing `EventsRelativeBox` for the days-to-go badge deletes the inline copy **and** fixes the always-red-here / grey-on-the-home-page inconsistency in the same move.

components/filter/events.tsx
hveraus: Confirmed and fixed in 174d3661.

The card is now a `ProductCardShell` with `href={event.url}`, so the anchor **is** the card — the overlay, the `z-10`/`z-20` layering and the `pointer-events-none` on the arrow are all gone. Verified the same way you did, on a running dev server: the focused element is the `<a>` itself, its parent is `overflow: visible`, and the ring paints in both themes.

On the 1px side note — you were right, and it needed a different fix than `outline-2`. tailwind-merge treats a bare `outline` and `outline-2` as one group and keeps only the last, so that pair can never set both a solid style and a 2px width. I measured it: with both classes present, `focus-visible:outline` is the one dropped, leaving you on the UA `outline-style: auto`.

So `cardShell` now uses `focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-brand`, matching `sidebarNavItem`'s existing choice. The `outline-none` is load-bearing: without it Chromium painted its own blue `:focus-visible` ring on top of the red one. Measured 2px brand ring, light and dark. Added `outline-none` to `sidebarNavItem` too, which had the same latent overlap.

components/filter/events.tsx
hveraus: Fixed by the same move to `ProductCardShell`, so a URL-less event renders a plain `div` instead of dropping the childless overlay entirely.

One thing the div fallback alone doesn't fix, which your comment does call out: the wrapper still lit up on hover with nothing behind it. So `hover:border-brand`, `hover:bg-card-hover` and the `active:*` pair are now gated on `Boolean(event.url)`, and so is the `ArrowCircle` — a card that isn't a link shouldn't show the "go" affordance either.

Didn't add a card-level test for it: `Event` isn't exported and pulls in `useFormatDates`, and `ProductCardShell`'s own fallback is already pinned by `__tests__/components/products/productCard.tsx`. Say the word if you'd rather have one anyway.

components/filter/events.tsx
hveraus: Partly actioned.

`EventMetaItem` and the days-to-go badge are now in `components/events/eventMeta.tsx`, imported by both this file and the v3 block. v3's local copies are deleted; its `DaysToGo` keeps the day maths and renders the shared chrome, and its three tina-annotated call sites now pass `tinaFieldValue={tinaField(event, ...)}` so the shared module stays free of `tinacms/dist/react`.

I didn't fold in `EventMetaGrid`, because the two aren't actually the same component — v3's is `grid grid-cols-1 sm:grid-cols-2` with a `className` passthrough, this one was a flex column with fixed colours. Inlined this one instead, per your other comment on line 345.

**Not actioned: the grey/red inconsistency**, and I think the framing there is slightly off. This badge matches v3's `DaysToGoBadge` character for character, and v3 is unconditionally red — that's the home block these cards were styled from, so they agree. The grey-unless-today one is `EventsRelativeBox`, which is used by the legacy `components/blocks/upcomingEvents.tsx`. Adopting it here would regress dark mode: it hardcodes `text-black` and `bg-gray-25` on the badge and `text-gray-500` on the date, none of it theme-aware. Happy to unify all three renderings, but it wants its own PR and a call on which behaviour wins.

components/presenters/presenterList.tsx
hveraus: Done — `__tests__/components/presenters/presenterList.tsx`, three cases:

- default: 2 anchors, `href` on each
- `linkless`: 0 anchors, names still present
- no presenter has a name: renders nothing

The linkless assertion reads `container.textContent` rather than `getByText`, because `linkless` renders the names as bare text nodes with no element of their own — `getByText` can't match them. Worth knowing if anyone extends these.

components/filter/FilterNav.tsx
hveraus: Both done.

`useEvents` now sets `label: "Technology"` and `label: "Format"` on the two groups, via a new `EventFilterGroup = FilterGroupProps & { label: string }` — kept off `FilterGroupProps` itself so `/clients` and `/opportunities` are untouched. `FilterNav` uses it for `aria-label`, so the landmarks read "Technology" and "Format" instead of "All Technology navigation".

`aria-current="location"` swapped for `aria-pressed`, agreed — these filter in place, they don't navigate. Verified in the rendered DOM: two `nav[aria-label]` with the new names, and `aria-pressed="true"` on the selected button in each group.

tailwind.config.js
hveraus: You're right, and the description was simply wrong. Rewritten.

It now says `styles.css` is untouched but `tailwind.config.js` is not, with the `sidebar` / `sidebar-narrow` retune spelled out and **"this resizes `/consulting` too"** stated explicitly. It also owns up to two tokens the old text missed entirely: `min-h-sidebar-card` (200px) and `max-w-sidebar-card` (300px) are both new, so "every token used already existed" was wrong as well.

The `min-h-11` → `min-h-9` point is in there too, as its own note: desktop rows go 44px → 36px, mobile unaffected because `py-2.5` already renders ~40px.

tina/collections/events.tsx
hveraus: Done, deleted outright rather than relabelled — schema field, the prop through `app/(events)/events/index.tsx` and `components/filter/events.tsx`, the type on `EventsFilterProps`, and the render block. `tina/tina-lock.json` regenerated off a dev server run.

One small correction to the reasoning: `afterEvents` isn't quite the *exact* same slot — `sidebarBody` rendered inside the content column of the grid, while `afterEvents` renders full-bleed below the page container. Doesn't change the conclusion though: the content was already empty, and two editor fields for "stuff after the events" is still one too many.

components/filter/events.tsx
hveraus: Done. The five values (`events`, `isLoading`, `hasMore`, `isFetching`, `fetchNext`) are picked with a ternary up front, then one `EventsList` + one `LoadMore`, with `key={pastSelected ? "past" : "upcoming"}` to keep the remount the two `TabPanel`s used to give for free. ~29 lines down to about 10.

Verified the toggle still behaves: 10 upcoming cards, click through, heading flips to "Past Events", 10 past cards.

components/filter/events.tsx
hveraus: Inlined both. The sharing on the comment above only ended up covering `EventMetaItem` and the badge — v3's `EventMetaGrid` is a two-column grid with a `className` passthrough, so there was no single component to share here. That leaves these two as exactly what you describe: one-div wrappers used once each.

components/events/eventsSidebar.tsx
hveraus: Done. `FilterNav` now lives at `components/filter/FilterNav.tsx` as the themed sibling of `FilterGroup`, `components/events/eventsSidebar.tsx` is deleted, and `components/filter/events.tsx` calls `StickySidebarLayout` directly. That also kills the import back across the boundary you spotted.

app/(events)/events/index.tsx
hveraus: Both actioned, the first a bit differently than suggested.

**The wrapper** is now `sidebarPageContainer`, exported from `components/layout/stickySidebar.tsx` and used by both pages, rather than folded into `StickySidebarLayout` itself — the container also wraps the breadcrumbs, which sit outside the layout component on `/events` and `/consulting` alike. Same outcome for your actual concern: the next padding tweak lands in one place.

**`EventTrimmed`** — `description` and `presenterProfileUrl` moved to `EventInfo` rather than being deleted. The type isn't quite down to one consumer: `EventInfo extends EventTrimmed`, and `app/live-steam-banner/liveStreamWidget.tsx` renders `event.description` through `TinaMarkdown` while `app/(events)/live` and the netug page both read `presenterProfileUrl`. Moving them keeps the index type narrow without breaking the live-banner path. Query fragment untouched, as you said.

content/events/index/index.mdx
hveraus: Done — all five deleted (`AzureOnDemand`, `AngularOnDemand`, `CAOnDemand`, `dotNet5OnDemand`, `AzureOnVimeo`, 180KB confirmed). Re-checked for references across `app/`, `components/`, `content/`, `tina/`, `helpers/` and `hooks/` first: none.

The removal is also now called out in the description, under a new "Deliberate removals" section, so it reads as intentional alongside the Vimeo content it belonged to.

hveraus:
Thanks Isaac — all 12 comments actioned in 174d3661, replies on each thread with the specifics. Summary:

**The blocking one is fixed and verified.** The card is a `ProductCardShell` with `href={event.url}`, so the anchor is the card and nothing clips its own ring. Your 1px side note needed a different fix than `outline-2`: tailwind-merge treats bare `outline` and `outline-2` as one group and keeps only the last, so `cardShell` moved to `focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-brand`, matching `sidebarNavItem`. Measured a 2px brand ring in both themes on a running dev server. `outline-none` turned out to be load-bearing — without it Chromium painted its own blue ring over the red one, and `sidebarNavItem` had the same latent overlap, so it got the same treatment.

**Done as asked:** the URL-less dead card (plus gating the hover chrome and the arrow on having a url, which the div fallback alone didn't cover), the `linkless` test, the two aria fixes, the collapsed timeframe branches, `FilterNav` extracted and `EventsSidebar` deleted, `sidebarBody` deleted outright, the shared page container, the five orphaned images, and the description.

**Three places I diverged**, each with the reasoning on its thread:

- `EventMetaGrid` isn't shared with v3 — v3's is a two-column grid with a `className` passthrough, not the same component. `EventMetaItem` and the badge are shared; the grid/row wrappers are inlined per your other comment.
- The grey-vs-red badge is **not** actioned. This badge matches v3's `DaysToGoBadge` exactly and v3 is unconditionally red; the grey one is `EventsRelativeBox` in the legacy `upcomingEvents.tsx`, which hardcodes `text-black` / `bg-gray-25` / `text-gray-500` and would regress dark mode here. Worth unifying, but as its own PR.
- `EventTrimmed`'s two fields moved to `EventInfo` rather than being deleted — `EventInfo extends EventTrimmed` and the live-stream banner renders `description` while `/live` and netug read `presenterProfileUrl`.

The description also now owns two things it previously got wrong beyond what you flagged: `min-h-sidebar-card` and `max-w-sidebar-card` are new tokens, so "every token used already existed" wasn't true either. The deliberate removals (Vimeo content, presenter/office links, `sidebarBody`) have their own section now.

`pnpm lint` and `pnpm test` pass (38 tests). Toggle, filters and the v3 home block smoke-tested in the browser; `tina-lock.json` regenerated for the schema change.

components/filter/events.tsx
```diff
@@ -330,85 +369,89 @@ const Event = ({ visible, event, jsonLd }: EventProps) => {
...
+                  </>
+                ) : null}
+              </EventMetaItem>
+              <EventMetaItem icon={FiMapPin}>{eventSite.name}</EventMetaItem>
+              <EventMetaItem icon={FiUser}>
+                {event.presenterName ? (
+                  event.presenterName
+                ) : event.presenterList?.length > 0 ? (
+                  <PresenterList presenters={event.presenterList} />
+                ) : null}
```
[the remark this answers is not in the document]
hveraus: Addressed, this was the same defect Isaac's blocking comment landed on from the other side.

`PresenterList` is called with `linkless` here, so it renders presenter names as plain text and emits no nested anchors at all. That removes the focusable-but-unclickable links you describe. The office/location link went the same way for the same reason.

Two things on top of that, in 174d3661:

- The overlay is gone entirely. The card is now a `ProductCardShell` with `href={event.url}`, so the anchor *is* the card rather than an `absolute inset-0` layer over it — there is no longer a stacking layer for anything to hide under. That also fixed a related bug: because the overlay sat inside a wrapper with `overflow-hidden`, its focus ring was clipped and never painted.
- `linkless` is now pinned by a test — `__tests__/components/presenters/presenterList.tsx` counts anchors with and without the prop, so removing it can't quietly reintroduce the regression.

hveraus:
> Re: [review #5164043701](https://github.com/SSWConsulting/SSW.Website/pull/5008#pullrequestreview-5164043701) — the `CHANGES_REQUESTED` summary

Two asks in that review body, both done in 174d3661.

**The blocking focus ring.** Confirmed exactly as you described it, and fixed. The anchor was `absolute inset-0` inside a wrapper carrying `overflow-hidden`, so `outline-offset-2` put the ring 2px outside the wrapper's padding box and it was clipped. The card is now a `ProductCardShell` with `href={event.url}` — the anchor *is* the card, and an element's own `overflow` never clips its own ring. Verified on a running dev server, not in isolation: focused element is the `<a>`, its parent computes `overflow: visible`, and the ring paints in light and dark.

Your side note about the 1px width was right too and needed its own fix. `outline` and `outline-2` are one tailwind-merge group, so that pair can never set both a solid style and a 2px width — with both present, `focus-visible:outline` is the class that gets dropped, leaving the UA `outline-style: auto`. `cardShell` therefore moved to `focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-brand`, matching `sidebarNavItem`'s existing choice. The `outline-none` is load-bearing: without it Chromium painted its own blue ring on top of the red one. `sidebarNavItem` had the same latent overlap and got the same treatment.

**Stating the deliberate removals.** Agreed, and the description now has a `Deliberate removals` section covering the Vimeo/On Demand content (plus the five images it was the only reference to), the presenter and office links with the reason they had to go, and `sidebarBody`. Detail replies are on each of the inline threads — note that most of them now read as *outdated* on the Files changed tab, since the fix commit rewrote those lines, so they sit behind **Show outdated**.

hveraus:
> Re: [review #5164116998](https://github.com/SSWConsulting/SSW.Website/pull/5008#pullrequestreview-5164116998) — the optional simplifications

All of these are actioned, and your framing held up: the whole thing came out roughly net neutral in lines while deleting a component, a schema field, a wrapper pair, five images and ~29 lines of branch duplication — **211 insertions, 201 deletions**, and 58 of those insertions are the new `PresenterList` test. `components/filter/events.tsx` alone is down 166 changed lines to a much shorter file.

You were right that the blocking fix pays for itself as a deletion: moving to `ProductCardShell` removed the overlay anchor, the `z-10`/`z-20` layering and the `pointer-events-none` on the arrow, and took the dead-card case with it.

**One item I did not take: reusing `EventsRelativeBox` for the badge.** The premise doesn't hold up as stated. This badge matches v3's `DaysToGoBadge` character for character, and v3 is *unconditionally* red — the v3 block is the home Upcoming Events section these cards were styled from, so the two already agree. The grey-unless-running/today rendering is `EventsRelativeBox`, which belongs to the legacy `components/blocks/upcomingEvents.tsx`. Dropping that in here would regress dark mode outright: it hardcodes `text-black` and `bg-gray-25` on the badge and `text-gray-500` on the date, none of it theme-aware, on a card that now has a dark surface.

What I did instead was remove the actual duplication you spotted: `EventMetaItem` and the badge chrome now live in `components/events/eventMeta.tsx` and are imported by both this file and the v3 block, so the identical class string exists once. Unifying the third rendering is worth doing and needs a call on which behaviour wins — happy to raise it separately if you want it tracked.

hveraus:

isaaclombardssw (approved):
LGTM

